PR #58, opened 2024-01-15
chore: log the rpc url when starting

mayconamaroCW:
added the rpc server url to the message log when it starts


dinhani-cw:
It would be useful to add the URL to the application configuration in `config.rs` and receive it as parameter.

mayconamaroCW:
@dinhani-cw done, please review 🙏

arthur-cw (approved):

PR #61, opened 2024-01-17
Postgres support pt3

arthur-cw:


src/eth/primitives/index.rs
```diff
@@ -0,0 +1,27 @@
+use sqlx::database::HasValueRef;
+use sqlx::error::BoxDynError;
+
+use crate::gen_newtype_from;
+
+pub struct Index(i32);
```
dinhani-cw, line 6: * Please add a doc explaining it is used in TransactionIndex and LogIndex.
* Use a unsigned integer because we cannot have a negative index.

src/eth/storage/mod.rs
```diff
@@ -3,7 +3,7 @@
 mod eth_storage;
 mod inmemory;
 mod metrified;
-mod postgres;
+pub mod postgres;
```
dinhani-cw, line 6: Is it really necessary everything from postgres public?

dinhani-cw (dismissed):

PR #66, opened 2024-01-18
Postgres integration pt4

arthur-cw:


src/eth/storage/postgres/types.rs
```diff
@@ -38,6 +53,18 @@ pub struct PostgresLogs {
     pub block_hash: Hash,
 }
 
+impl PostgresLog {
+    fn into_log_mined(&self) -> LogMined {
```
dinhani-cw, line 52: I think `impl From<PostgresLog> for LogMined` can be used here.

src/eth/miner/block_miner.rs
```diff
@@ -63,15 +65,15 @@ impl BlockMiner {
                 let mined_log = LogMined {
                     log: mined_log,
                     transaction_hash: input.hash.clone(),
-                    transaction_index,
-                    log_index,
+                    transaction_index: transaction_index.clone(),
```
dinhani-cw, line 68: Derive `Copy` for `Index`.

src/eth/miner/block_miner.rs
```diff
@@ -63,15 +65,15 @@ impl BlockMiner {
...
+                    log_index: log_index.clone(),
                     block_number: block.header.number,
                     block_hash: block.header.hash.clone(),
                 };
                 mined_logs.push(mined_log);
 
                 // increment log index
-                log_index += 1;
+                log_index = log_index + Index::new(1);
```
dinhani-cw, line 76: Derive `[Add](https://docs.rs/derive_more/0.99.17/derive_more/#operators)` for `Index`.

src/eth/primitives/index.rs
```diff
@@ -1,17 +1,41 @@
...
+use ethereum_types::U256;
+use ethereum_types::U64;
 use sqlx::database::HasValueRef;
 use sqlx::error::BoxDynError;
 
 use crate::gen_newtype_from;
 
-#[allow(dead_code)]
-/// Representation of a transaction index or log index.
+/// Index is a representation for transaction indexes and log indexes
```
dinhani-cw, line 10: Some practices from the Rust community when writing docs:
* No need to repeat the struct name or have phrases like "This is struct" or "This method" because documentation is contextual.
* Also all phrases should ends with a period.
* Verbs are always written as third-person singular.

Confirm this by checking the documentation of popular repositories.

src/eth/primitives/index.rs
```diff
@@ -1,17 +1,41 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, fake::Dummy, serde::Serialize, serde::Deserialize)]
 pub struct Index(u16);
 
+impl Index {
+    pub fn new(inner: u16) -> Self {
+        Index(inner)
+    }
+}
+
+impl Add for Index {
```
dinhani-cw, line 20: Use `derive_more`.

src/eth/primitives/index.rs
```diff
@@ -1,17 +1,41 @@
...
+}
+
 // -----------------------------------------------------------------------------
 // Conversions: Other -> Self
 // -----------------------------------------------------------------------------
 gen_newtype_from!(self = Index, other = u16);
 
+impl From<i32> for Index {
+    fn from(value: i32) -> Self {
+        value.into()
```
dinhani-cw, line 35: I believe this is a bug. It will call the function recursively.

src/eth/primitives/rs.rs
```diff
@@ -0,0 +1,36 @@
+use ethereum_types::U256;
+use sqlx::database::HasValueRef;
+use sqlx::error::BoxDynError;
+
+use crate::gen_newtype_from;
+
+// Type representing `r` and `s` variables
+// from the ECDSA signature
+pub struct Rs(U256);
```
dinhani-cw, line 9: Consider if naming it can be more descriptive `EcdsaRs`.

src/eth/primitives/v.rs
```diff
@@ -0,0 +1,36 @@
+use ethereum_types::U64;
+use sqlx::database::HasValueRef;
+use sqlx::error::BoxDynError;
+
+use crate::gen_newtype_from;
+
+// Type representing `v` variable
+// from the ECDSA signature
+pub struct V(U64);
```
dinhani-cw, line 9: Consider if naming it `EcdsaV` can be more descriptive.

src/eth/primitives/rs.rs
arthur-cw: makes sense

src/eth/miner/block_miner.rs
arthur-cw: I chose not to derive copy and explicitly call `clone()` so we can later optimize those.

arthur-cw: but since Index is just an integer, I think it's OK to derive Copy now.

static/schema/001-init.sql
```diff
@@ -33,12 +33,20 @@ CREATE TABLE IF NOT EXISTS transactions (
...
-    gas NUMERIC NOT NULL CHECK (gas >= 0),
-    idx_in_block SERIAL NOT NULL CHECK (idx_in_block >= 0) UNIQUE,
-    block_number BIGSERIAL REFERENCES blocks(number) NOT NULL CHECK (block_number >= 0),
-    block_hash BYTEA REFERENCES blocks(hash) NOT NULL CHECK (LENGTH(block_hash) = 32),
-    PRIMARY KEY (hash)
+    input BYTEA NOT NULL CHECK (LENGTH(input) <= 24000)
+    ,output BYTEA NOT NULL
+    ,gas NUMERIC NOT NULL CHECK (gas >= 0)
+    ,gas_price NUMERIC NOT NULL CHECK (gas_price >= 0)
+    ,idx_in_block SERIAL NOT NULL CHECK (idx_in_block >= 0) UNIQUE
```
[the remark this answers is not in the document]
arthur-cw: Yeah, in fact there can be many indexes. The problem here is that I want to reference this row from other tables. I will see what can be done here.

src/eth/storage/postgres/types.rs
```diff
@@ -14,12 +26,60 @@ pub struct PostgresTransaction {
...
+    pub value: Wei,
+    pub r: Rs,
+    pub s: Rs,
+    pub v: V,
 }
 
-pub struct PostgresLogs {
+impl PostgresTransaction {
+    pub fn into_transaction_mined(self, logs: Vec<PostgresLog>, topics: Vec<PostgresTopic>) -> TransactionMined {
+        let mined_logs: Vec<LogMined> = logs.iter().map(|log| log.into_log_mined(topics.clone())).collect();
```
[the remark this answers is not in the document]
arthur-cw: >Are all mined_logs supposed to contain all topics?

Yes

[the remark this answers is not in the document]
arthur-cw: > Also there is a lot of cloning going on
Yes, I did this the more dirty way so I can optimize in the end, that will be fixed soon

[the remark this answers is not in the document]
arthur-cw: > maybe you could change PostgresLog::into_log_mined to take self instead of &self and consume the log when doing the conversion, then you could use into_iter() here. You're dropping logs anyway.

Yes, I tried that but ran into other errors and decided to use clone just to have everything working before coming back to optimize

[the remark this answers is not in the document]
carneiro-cw: > > Are all mined_logs supposed to contain all topics?
> 
> Yes, a topic is what basically defines the log

I understand, but arent you adding all topics to all logs? Wouldn't that make all of the logs in a transaction the same since they are basically defined by the topics and each of the logs contain all of the same topics?

src/eth/primitives/index.rs
arthur-cw: thanks!

dinhani-cw (dismissed):

Issue #26, opened 2023-12-22
Remove `thiserror` crate and use `anyhow`


arthur-cw:

renancloudwalk:
👍

carneiro-cw:
Maybe we can use them both, if an EthError (or any other error) does not provide enought context just call `.with_context(|| "context")`.

arthur-cw:
@carneiro-cw agreed

PR #71, opened 2024-01-24
feat: Remove eyre and thiserror in favor of anyhow

carneiro-cw:
Addresses #26.


src/eth/rpc/rpc_parser.rs
```diff
@@ -54,8 +60,28 @@ pub fn rpc_parsing_error<S: serde::Serialize>(message: S) -> ErrorObjectOwned {
     ErrorObjectOwned::owned(PARSE_ERROR_CODE, PARSE_ERROR_MSG, Some(message))
 }
 
-impl From<EthError> for ErrorObjectOwned {
-    fn from(error: EthError) -> Self {
-        rpc_internal_error(error.to_string())
+pub struct RPCError(anyhow::Error);
```
dinhani-cw, line 63: `RpcError`

src/eth/rpc/rpc_parser.rs
```diff
@@ -54,8 +60,28 @@ pub fn rpc_parsing_error<S: serde::Serialize>(message: S) -> ErrorObjectOwned {
...
+impl Display for RPCError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Display::fmt(&self.0, f)
+    }
+}
+impl Debug for RPCError {
```
dinhani-cw, line 69: Can you just derive `Debug` instead?

src/eth/rpc/rpc_parser.rs
```diff
@@ -54,8 +60,28 @@ pub fn rpc_parsing_error<S: serde::Serialize>(message: S) -> ErrorObjectOwned {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Debug::fmt(&self, f)
     }
 }
+
+impl From<anyhow::Error> for RPCError {
```
dinhani-cw, line 75: Check the [`gen_newtype_from`](https://github.com/cloudwalk/stratus/blob/93e35fa06d5497afb2156b357295b352173aae9e/src/ext.rs#L7) macro to derive `From` macros based on the inner type.

Example:
`gen_newtype_from!(self = RpcError, other = anyhow::Error, OtherErrorThatConvertToAnyhowError)`

src/eth/rpc/rpc_server.rs
```diff
@@ -15,6 +16,8 @@ use jsonrpsee::IntoSubscriptionCloseResponse;
 use jsonrpsee::PendingSubscriptionSink;
 use serde_json::Value as JsonValue;
 
+use super::rpc_internal_error;
```
dinhani-cw, line 19: We are usually using the full name qualifier for imports.

dinhani-cw:
The feature itself is fine. Just added some patterns we are trying to follow and maybe can also be automated with a linter.

src/eth/rpc/rpc_parser.rs
carneiro-cw: Ah I see, I was not aware of the acronym naming convention. The revm library even has structures that don't follow this convention such as `revm::EVMData`. I'll fix this tommorow.

src/eth/rpc/rpc_parser.rs
carneiro-cw: Well, kinda. If I derive Debug formatting `RPCError` will produce something like this: `"RPCError(inner.fmt())"`, but what I wanted was for RPCError to behave exactly like an anyhow::Error (although I didn't implement Deref, because I wasn't sure we'd need it and I did not want to bloat this struct). I can change it If you think it'd be better.

src/eth/rpc/rpc_parser.rs
carneiro-cw: I'll take a look tomorrow.

src/eth/rpc/rpc_server.rs
carneiro-cw: Sorry, auto-import

src/eth/rpc/rpc_parser.rs
dinhani-cw: For `Display` trait, it is fine to delegate to the inner implementation because the goal is an human-friendly message. But for `Debug`, we should stick with the struct representation when possible.

It is the same idea of `repr` and `str` from Python.

src/eth/rpc/rpc_parser.rs
dinhani-cw: Not critical, but it just make things easier when we have this pattern to delegate to the wrapped type implementation.

src/eth/rpc/rpc_parser.rs
dinhani-cw: Take jsonrpsee as reference:
https://docs.rs/jsonrpsee/0.21.0/jsonrpsee/?search=rpc

dinhani-cw:
Execute `just lint` to have all an overview of all warnings.
There are unused imports, unused variable and unreachable code.

dinhani-cw (approved):

PR #72, opened 2024-01-24
feat: subscriptions improvements

dinhani-cw:
* Using Subscription ID as String instead of number to be compatible with Geth client.
* Fix log filter:
  * It can receive multiple address.
  * It can received multiple combination of topics.

### DIFF
--- a/src/eth/primitives/log_filter_input.rs
+++ b/src/eth/primitives/log_filter_input.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use itertools::Itertools;
 use serde_with::formats::PreferMany;
 use serde_with::serde_as;
 use serde_with::OneOrMany;
@@ -8,6 +9,7 @@ use crate::eth::primitives::Address;
 use crate::eth::primitives::BlockSelection;
 use crate::eth::primitives::Hash;
 use crate::eth::primitives::LogFilter;
+use crate::eth::primitives::LogFilterTopicCombination;
 use crate::eth::primitives::LogTopic;
 use crate::eth::primitives::StoragePointInTime;
 use crate::eth::storage::EthStorage;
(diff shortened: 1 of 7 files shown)


src/eth/primitives/log_filter_input.rs
```diff
@@ -21,11 +28,13 @@ pub struct LogFilterInput {
...
 
-    #[serde(default)]
-    pub address: Option<Address>,
+    #[serde(rename = "address", default)]
+    #[serde_as(deserialize_as = "OneOrMany<_, PreferMany>")]
+    pub address: Vec<Address>,
 
-    #[serde(default)]
-    pub topics: Vec<Option<LogTopic>>,
+    #[serde(rename = "topics", default)]
```
carneiro-cw, line 35: Another redundant rename

src/eth/primitives/log_filter_input.rs
```diff
@@ -21,11 +28,13 @@ pub struct LogFilterInput {
     #[serde(rename = "blockHash", default)]
     pub block_hash: Option<Hash>,
 
-    #[serde(default)]
-    pub address: Option<Address>,
+    #[serde(rename = "address", default)]
```
carneiro-cw, line 31: Isn't the renaming redundant here?

carneiro-cw (commented):
Only thing I noticed out of the ordinary was those two redundant renames. I imagine there is a reason for it since previously they were not present.

dinhani-cw:
@carneiro-cw 
The two new renames are not strictly necessary. 

They are an additional protection against someone renaming one of these fields without knowing they are part of an specification and breaking things unknowingly.

The two modified fields became `Vec` in this PR, so someone may be tempted to rename `address` to `addresses`. Even if the field is renamed, the JSON parsing will not break.

src/eth/primitives/log_filter_input.rs
dinhani-cw: Answer below.

src/eth/primitives/log_filter_input.rs
dinhani-cw: See below.

PR #74, opened 2024-01-24
feat: Migrate to async traits for EthStorage

carneiro-cw:
Note that in the file `src/eth/executor.rs` in the functions `EthExecutor::transact` and `EthExecutor::call`, I aquire the `evm_lock`, do nothing with it, and then create a new `Revm` object and pass that to `spawn_blocking`. There are two ways that I thought of to circumvent this, but I'd like further discussion before implementing either of them.

1. Run the EVM in a dedicated thread (instead of spawning a thread for each call) with some form of communication, maybe a channel or queue.
2. Require `Clone` for `EthStorage` and `Evm`. Then I could aquire the lock and clone the underlying struct. However implementing clone for dyn traits is not trivial, to solve this there is a popular crate that includes a macro for it (https://github.com/dtolnay/dyn-clone).

Are either of these approaches desirable at this time?

### DIFF
--- a/src/eth/storage/eth_storage.rs
+++ b/src/eth/storage/eth_storage.rs
@@ -1,5 +1,5 @@
 use anyhow::anyhow;
-use jsonrpsee::core::async_trait;
+use async_trait::async_trait;
 
 use crate::eth::primitives::Account;
 use crate::eth::primitives::Address;
--- a/src/eth/storage/eth_storage.rs
+++ b/src/eth/storage/eth_storage.rs
@@ -1,4 +1,5 @@
 use anyhow::anyhow;
+use jsonrpsee::core::async_trait;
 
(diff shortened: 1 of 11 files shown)


src/eth/storage/eth_storage.rs
```diff
@@ -1,4 +1,5 @@
 use anyhow::anyhow;
+use jsonrpsee::core::async_trait;
```
dinhani-cw, line 2: You are using a reexport from jsonrpsee instead of the original create.

dinhani-cw:
@carneiro-cw 
The EVM approach is desirable, but I am already working on it.
For this PR what you did is enough.
The final approach will use mpmc channels so we can have parallel execution if desirable.

arthur-cw (approved):

PR #75, opened 2024-01-24
Eth documentation

renancloudwalk:
## Summary

This pull request introduces enhanced documentation across several modules of the Stratus project, with a specific focus on elucidating the non-time-dependent nature of our block generation process. The updates include detailed descriptions of key functions and structures, particularly in the `EthExecutor` and `BlockMiner` components, which play crucial roles in our unique blockchain implementation.

## Key Changes

- Updated `EthExecutor::transact` method documentation to highlight the non-time-dependent block generation after transaction execution.
- Refined `BlockMiner` documentation in `miner.rs` to emphasize its role in processing transactions into blocks outside the conventional time-based paradigm.
- Enhanced readability and provided clearer explanations in `executor.rs` and `miner/mod.rs`, ensuring that the purpose and functioning of these modules are well-understood.

## Rationale

The Stratus project's deviation from traditional time-based block generation to a transaction-processing-based approach is a cornerstone of our blockchain architecture. These documentation updates aim to make this distinction clear, aiding both current and future developers in understanding and contributing effectively to our codebase.

## Additional Notes

- Some TODOs have been added for potential future optimizations and refactoring considerations.
- Further documentation enhancements may be required as the project evolves, especially in areas interfacing directly with the block generation logic.

Your review and input on these changes would be greatly appreciated to ensure clarity and completeness.

### DIFF
--- a/src/eth/executor.rs
+++ b/src/eth/executor.rs
@@ -50,14 +50,13 @@ impl EthExecutor {
         }
     }
 
-    /// Executes a transaction and produces a new block.
+    /// Executes Ethereum transactions and facilitates block creation.
     ///
-    /// This function handles the entire lifecycle of a transaction. It starts by validating
-    /// the transaction, then executes it using the EVM, and finally, mines a new block using
-    /// `mine_with_one_transaction`, irrespective of time-based criteria. This approach aligns
-    /// with the Stratus project's strategy of block generation based on transaction processing
-    /// rather than fixed time intervals. The function also manages state mutations and broadcasts
-    /// notifications for new blocks and logs.
+    /// This function is a key part of the transaction processing pipeline. It begins by validating
+    /// incoming transactions and then proceeds to execute them. Unlike conventional blockchain systems,
+    /// the block creation here is not dictated by timed intervals but is instead triggered by transaction
+    /// processing itself. This method encapsulates the execution, block mining, and state mutation,
+    /// concluding with broadcasting necessary notifications for the newly created block and associated transaction logs.
     ///
     /// TODO: Optimize the cloning operations to enhance performance.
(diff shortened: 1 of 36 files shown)


src/eth/executor.rs
```diff
@@ -14,12 +21,17 @@ use crate::eth::primitives::TransactionExecution;
 use crate::eth::primitives::TransactionInput;
 use crate::eth::storage::EthStorage;
 
-/// High-level coordinator of Ethereum transactions.
+/// The EthExecutor struct is responsible for orchestrating the execution of Ethereum transactions.
```
dinhani-cw, line 24: This EthExecutor internal working will change soon. It will use multi-producer multi-consumer (mpmc) channels with a pool of EVMs, so we can:

* avoid locks
* allow multiple transactions to execute in parallel

But for now it is ok because it is describing what we have here.

src/eth/executor.rs
```diff
@@ -38,9 +50,15 @@ impl EthExecutor {
         }
     }
 
-    /// Execute a transaction, mutate the state and return function output.
+    /// Executes Ethereum transactions and facilitates block creation.
```
dinhani-cw, line 53: It would be useful to have similar documentation for the `call` function because it is used in `eth_call` and `eth_estimateGas` operations.

dinhani-cw:
Very good work.

A few points to improve based in the premise we want to have the docs deployed somewhere:

* Execute `just doc` and check how the docs look when viewed from the Rust doc page.
  * There are a few places that needs a blank line between the header and the full description.
* When possible, add cross-reference link to mentioned structs and functions.

src/eth/executor.rs
renancloudwalk: mpmc will be awesome

dinhani-cw (approved):

PR #76, opened 2024-01-25
feat: execute transaction in parallel evms

dinhani-cw:


src/eth/rpc/rpc_subscriptions.rs
```diff
@@ -33,16 +33,14 @@ pub struct RpcSubscriptions {
...
-        tokio::spawn(async move {
-            loop {
-                for closed_sub in self.new_heads.iter().filter(|sub| sub.is_closed()) {
-                    self.new_heads.remove(closed_sub.key());
-                }
-                for closed_sub in self.logs.iter().filter(|sub| sub.0.is_closed()) {
-                    self.logs.remove(closed_sub.key());
-                }
-                sleep(CLEANING_FREQUENCY).await;
+        tokio::task::spawn_blocking(move || loop {
```
dinhani-cw, line 36: Use spawn_blocking instead of spawn because Dashmap was causing deadlocks when running in the async context.

src/main.rs
```diff
@@ -25,11 +32,20 @@ async fn main() -> anyhow::Result<()> {
...
 
     // start rpc server
     serve_rpc(executor, storage, config.address).await?;
-
     Ok(())
 }
+
+fn init_evms(storage: Arc<dyn EthStorage>) -> NonEmpty<Box<dyn Evm>> {
+    let mut evms: Vec<Box<dyn Evm>> = Vec::with_capacity(10);
+    for _ in 1..=10 {
```
dinhani-cw, line 47: The number of EVMS should be configurable.

src/main.rs
```diff
@@ -1,18 +1,25 @@
...
 use stratus::eth::storage::InMemoryStorage;
 use stratus::eth::EthExecutor;
+use stratus::ext::new_tokio_runtime;
 use stratus::infra;
 use stratus::infra::postgres::Postgres;
 
-#[tokio::main]
-async fn main() -> anyhow::Result<()> {
+fn main() -> anyhow::Result<()> {
+    let runtime = new_tokio_runtime("tokio-main", 1, 1);
```
dinhani-cw, line 18: The number of async and blocking threads should be configurable.

It is being set as 1 for now because it makes easier to detect deadlocks and other mistakes in async code.

src/eth/storage/inmemory.rs
```diff
@@ -3,10 +3,12 @@
 use std::collections::HashMap;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
-use std::sync::RwLock;
 
 use async_trait::async_trait;
 use indexmap::IndexMap;
+use tokio::sync::RwLock;
```
dinhani-cw, line 9: Use tokio sync primitives to avoid deadlocks.

src/eth/executor.rs
```diff
@@ -21,29 +24,42 @@ use crate::eth::primitives::StoragePointInTime;
...
+type EvmTask = (EvmInput, oneshot::Sender<anyhow::Result<TransactionExecution>>);
 
 /// The EthExecutor struct is responsible for orchestrating the execution of Ethereum transactions.
 /// It holds references to the EVM, block miner, and storage, managing the overall process of
 /// transaction execution, block production, and state management.
 pub struct EthExecutor {
-    // Mutex-wrapped EVM for synchronized access to EVM operations.
-    evm: Mutex<Box<dyn Evm>>,
+    // Tokio runtime where background EVMs are executed (keep reference because it cannot be dropped).
+    _runtime_guard: Runtime,
```
carneiro-cw, line 39: In the comment abote you say we should keep a reference to Runtime because it cannot be dropped. But you're keeping the value, is this intentional? Did I misunderstand?

carneiro-cw (commented):
Besides my one comment, LGTM.

PR #79, opened 2024-01-26
feat: use rayon threadpool to run evms

dinhani-cw:
* Tokio nested runtimes are not allowed, so use a Rayon thread-pool instead.
* Use crossbeam-channel instead of async-channel because Rayon is not async.
* Use `futures::executor::block_on` instead of `tokio::block_on` inside EVM because Tokio runtime is not available for EVM execution. It just blocks on the current thread, that is the desired behavior.
* Stop EVM threads when task channel is closed.

### DIFF
--- a/src/ext.rs
+++ b/src/ext.rs
@@ -4,11 +4,6 @@
 // Macros
 // -----------------------------------------------------------------------------
 
-use std::time::Duration;
-
-use tokio::runtime::Builder;
-use tokio::runtime::Runtime;
-
 /// Generates [`From`] implementation for a [newtype](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) that delegates to the inner type [`From`].
 #[macro_export]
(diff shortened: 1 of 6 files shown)


src/ext.rs
```diff
@@ -50,22 +45,6 @@ macro_rules! if_else {
     };
 }
 
-// -----------------------------------------------------------------------------
-// Tokio
-// -----------------------------------------------------------------------------
-
-/// Creates a new Tokio runtime with the specified configuration.
-pub fn new_tokio_runtime(thread_name: &'static str, num_async_threads: usize, num_blocking_threads: usize) -> Runtime {
```
dinhani-cw, line 58: Removed it from here because we should not create multiple Tokio runtimes, so the reuse does not make sense.

carneiro-cw (approved):

carneiro-cw (approved):
LGTM

PR #93, opened 2024-01-29
fix: type conversions causing overflows because of infinite recursions and save transaction changes only in case of success

carneiro-cw:


src/eth/primitives/nonce.rs
```diff
@@ -40,7 +40,7 @@ gen_newtype_from!(self = Nonce, other = u8, u16, u32, u64, u128, U256, usize, i3
 
 impl From<BigDecimal> for Nonce {
     fn from(value: BigDecimal) -> Self {
-        value.into()
+        Nonce(U256::from_dec_str(&value.to_string()).unwrap_or(U256::zero()))
```
renancloudwalk, line 43: Maybe I would lof this kind of failure, down the road it can make our life easier

renancloudwalk (approved):
Overall I don't like plain unwrap_or without logs.

src/eth/primitives/nonce.rs
carneiro-cw: Sure, I can log the unwrap_ors

renancloudwalk (dismissed):

dinhani-cw (dismissed):

arthur-cw:
@carneiro-cw just add some coments to the unwrap_or explaining why it's OK, and it's good

arthur-cw (approved):

PR #94, opened 2024-01-29
Cancel signal on rpc server

renancloudwalk:


carneiro-cw (dismissed):
LGTM

carneiro-cw:
There are a few lint errors though, in src/main.rs:112:5 you can remove the loop, src/main.rs:79:46 can remove the mut from cancel_signal, and src/main.rs:85:67 you can pass url without a reference.

PR #109, opened 2024-01-30
feat: add save slots in save block

carneiro-cw:


src/eth/storage/postgres/postgres.rs
```diff
@@ -431,6 +431,18 @@ impl EthStorage for Postgres {
                     .execute(&self.connection_pool)
                     .await
                     .context("failed to insert account")?;
+                    for (slot_idx, value) in change.slots {
+                        sqlx::query_file!(
+                            "src/eth/storage/postgres/queries/insert_account_slot.sql",
+                            &<[u8; 32]>::from(slot_idx),
+                            &<[u8; 32]>::from(value.take().ok_or(anyhow::anyhow!("no change for slot"))?.value),
```
arthur-cw, line 438: I'm not sure we should revert here if there are not changes. This should be captured earlier

carneiro-cw: It should be impossible for a slot change to get there with no changes. See `revm.rs > parse_revm_state`

arthur-cw (approved):

PR #115, opened 2024-01-30
feat: add parent_hash to block headers

mayconamaroCW:
### Changes

- Change block hash from random to the hash of its number so it is deterministic
- Add `parent_hash` as the hash of the previous number
- Add `parent_hash` to database schema

Added unit tests, tested manually with Insomnia


arthur-cw:
@mayconamaroCW let's use big endianess

arthur-cw (approved):
👍

PR #121, opened 2024-01-31
feat: debug_setHead

dinhani-cw:
Implements `debug_setHead`. It allows resetting the blockchain entire state to a past block.

Like other functions like `test_accounts`, this method should be put behind a future-flag in the future to avoid destructive actions in production.


renancloudwalk (commented):
I would just like a safety way to prevent this from clearing production :D

PR #125, opened 2024-01-31
feat: Postgres::reset

carneiro-cw:


arthur-cw (approved):
I'm not sure we should delete everything ON CASCADE, in contrast to deleting one by one, but I think it works now and it's good enough

PR #135, opened 2024-02-01
add Option to `From` address in `eth_call`

arthur-cw:


src/eth/evm/evm.rs
```diff
@@ -28,8 +28,9 @@ pub struct EvmInput {
     ///
     /// It can be:
     /// * Transaction signer when executing an `eth_sendRawTransaction`.
-    /// * Address present in the `from` field when performing an `eth_call`.
-    pub from: Address,
+    /// * Placeholder when performing an `eth_call`
+    /// * Not specified when performing an `eth_call`
+    pub from: Option<Address>,
```
dinhani-cw, line 33: EVM always assume a `from` when not specified. It defaults to zero.
So at least here it should be mandatory and the caller decide what value to use when an address is not available.
For transactions it will always be available because the signer will be used.

dinhani (approved):

PR #138, opened 2024-02-02
feat: postgres change table structure

carneiro-cw:


renancloudwalk (changes requested):
Stop the code!
I already have some changes on the way here.

carneiro-cw:
The naming was an accident

PR #140, opened 2024-02-02
feat: task to download blocks

dinhani-cw:


src/bin/task-get-blocks.rs
```diff
@@ -0,0 +1,140 @@
+use std::cmp::min;
```
renancloudwalk, line 1: I think this filename better be on snake_case.
https://rust-lang.github.io/api-guidelines/naming.html

PR #148, opened 2024-02-05
feat: save external blocks to postgres

dinhani-cw:


src/bin/importer/fetch_rpc.rs
```diff
@@ -0,0 +1,154 @@
...
+        Err(sqlx::Error::Database(e)) if e.is_unique_violation() => {
+            tracing::warn!(reason = ?e, "block unique violation, skipping");
+            Ok(())
+        }
+        Err(e) => {
+            tracing::error!(reason = ?e, "failed to insert block");
+            Err(e).context("failed to insert block")
+        }
+    }
+}
```
renancloudwalk, line 154: cant this be on a lib we both will have to call this

PR #149, opened 2024-02-05
fix: don't insert gensis block twice

carneiro-cw:


dinhani-cw (approved):

renancloudwalk (commented):
This is not enough to safeguard things, why not use a constraint as well?

carneiro-cw:
Can you elaborate? What is not being safeguarded? Which constraint on what? The objective of this PR was not to safeguard anything, it was just so Stratus could be restarted without restarting the db, since the database rightfully does not allow for two genesis blocks to be inserted.

PR #150, opened 2024-02-05
chore: add necessary methods to fetch slots and transaction details

renancloudwalk:


src/infra/blockchain_client.rs
```diff
@@ -53,4 +56,39 @@ impl BlockchainClient {
             }
         }
     }
+
+    pub async fn get_transaction_receipt(&self, transaction_hash: Hash) -> anyhow::Result<JsonValue> {
+        tracing::debug!(%transaction_hash, "retrieving block");
```
dinhani-cw, line 61: wrong log message

PR #153, opened 2024-02-06
chore: add env to RPC context

renancloudwalk:
it will be later used to create safety measures against deletions

### DIFF
--- a/src/eth/rpc/rpc_context.rs
+++ b/src/eth/rpc/rpc_context.rs
@@ -4,13 +4,7 @@ use std::sync::Arc;
 use crate::eth::rpc::RpcSubscriptions;
 use crate::eth::storage::EthStorage;
 use crate::eth::EthExecutor;
-
-pub enum Environment {
-    Development,
-    Test,
-    Staging,
-    Production,
-}
+use crate::config::Environment;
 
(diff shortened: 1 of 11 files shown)


src/eth/rpc/rpc_context.rs
```diff
@@ -5,6 +5,13 @@ use crate::eth::rpc::RpcSubscriptions;
 use crate::eth::storage::EthStorage;
 use crate::eth::EthExecutor;
```
dinhani-cw, line 7: I think this enum should be in the config module.

src/eth/rpc/rpc_context.rs
```diff
@@ -17,14 +24,27 @@ pub struct RpcContext {
...
             .field("chain_id", &self.chain_id)
             .field("client_version", &self.client_version)
+            .field("environment", &self.environment)
             .field("gas_price", &self.gas_price)
             .finish_non_exhaustive()
     }
 }
+
+impl Debug for Environment {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
dinhani-cw, line 42: Derive Debug?

PR #157, opened 2024-02-06
Add env to rpc context

renancloudwalk:
Passing config around to context, so we gather further information.
Also checking for current environment


arthur-cw:
lint

PR #158, opened 2024-02-06
Creates Env based conditions

renancloudwalk:
Passing config around to context, so we gather further information.
Also checking for current environment


renancloudwalk:
Previous PR had a weird conflict with the branch

PR #159, opened 2024-02-06
feat: check conflicts on postgres

carneiro-cw:


src/eth/storage/postgres/postgres.rs
```diff
@@ -471,23 +484,35 @@ impl EthStorage for Postgres {
...
-                        block_number
+                        block_number,
+                        original_nonce,
+                        original_balance
                     )
                     .execute(&mut *tx)
                     .await
                     .context("failed to insert account")?;
 
+                    if account_result.rows_affected() != 1 {
```
arthur-cw, line 506: can you create a function to represent this case? or at least add some comments

src/eth/primitives/execution_conflict.rs
```diff
@@ -58,4 +67,16 @@ pub enum ExecutionConflict {
...
+    Account {
+        address: Address,
+        expected_nonce: BigDecimal,
+        expected_balance: BigDecimal,
+    },
+
+    PgSlot {
+        address: Address,
+        slot: Vec<u8>,
+        expected: Vec<u8>,
```
arthur-cw, line 80: why not use [u8; 32] if it's gonna be converted later?

src/eth/storage/postgres/postgres.rs
```diff
@@ -508,25 +533,35 @@ impl EthStorage for Postgres {
...
                             change.address.as_ref(),
-                            block_number
+                            block_number,
+                            original_value
                         )
                         .execute(&mut *tx)
                         .await
                         .context("failed to insert slot")?;
 
+                        if slot_result.rows_affected() != 1 {
```
arthur-cw, line 555: same as above

src/eth/storage/postgres/postgres.rs
carneiro-cw: Sure

src/eth/primitives/execution_conflict.rs
carneiro-cw: True

arthur-cw (approved):

PR #172, opened 2024-02-08
chore: bypass chain id validation

renancloudwalk:
Otherwise we have a panic when chain id is not set. Stopping all the process

### DIFF
--- a/src/eth/primitives/transaction_input.rs
+++ b/src/eth/primitives/transaction_input.rs
@@ -114,8 +114,9 @@ impl TryFrom<EthersTransaction> for TransactionInput {
         let chain_id: ChainId = match value.chain_id {
             Some(chain_id) => chain_id.into(),
             None => {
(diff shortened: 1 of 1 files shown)


src/eth/primitives/transaction_input.rs
```diff
@@ -114,8 +114,9 @@ impl TryFrom<EthersTransaction> for TransactionInput {
         let chain_id: ChainId = match value.chain_id {
             Some(chain_id) => chain_id.into(),
             None => {
-                tracing::warn!(reason = %"transaction without chain id");
-                return Err(ConversionError::NoChainId(anyhow!("Transaction sent without chain id is not allowed.")));
+                let transaction_value = value.clone();
+                tracing::warn!(reason = %"transaction without chain id", ?transaction_value);
+                2008.into()
```
arthur-cw, line 119: are you testing with Testnet data? Because our Mainnet's chain id is actually 2009

PR #173, opened 2024-02-08
feat: implement read_account_opt and read_slot_opt

dinhani-cw:


src/bin/rpc_server_poller.rs
```diff
@@ -1,20 +1,12 @@
-use std::env;
```
dinhani-cw, line 1: removing unused imports here to pass all checks

arthur-cw (approved):
nit: I'd use the `maybe_` prefix for these methods, I really like this pattern.
otherwise it's OK

PR #191, opened 2024-02-09
feat: check block fields while importing

carneiro-cw:


src/eth/executor.rs
```diff
@@ -223,6 +265,26 @@ impl EthExecutor {
     }
 }
 
+fn cmp_execution_with_receipt(execution: &Execution, receipt: &ExternalReceipt) {
```
dinhani-cw, line 268: The purpose of the executor is just to orchestrate different components (EVM, Miner, Storage, etc).

I think it should be easier to maintain and test if we isolate the checks in some struct.

Example:
ExecutionChanges receives ExternalReceipt and compares current state with expected receipt state and logs.
ExecutionChanges receives ExternalBlock and compare other necessary stuff.

carneiro-cw: Extracted the comparison functions to Block and Execution. As a side-effect I had to add a clone [here](https://github.com/cloudwalk/stratus/blob/89c7fc0ad9b938a08d71b43a0ac657cab4bc9c9b/src/eth/executor.rs#L103)

src/eth/executor.rs
```diff
@@ -72,42 +71,77 @@ impl EthExecutor {
...
+            let execution = self.execute_in_evm(input).await?;
+
+            execution.cmp_with_receipt(tx_receipt);
+
+            let transaction_input = transaction.try_into().or(Err(anyhow!("failed to convert tx input")))?;
+            executions.push((transaction_input, execution));
         }
 
+        let block = if let Some(executions) = NonEmpty::from_vec(executions) {
+            self.miner.lock().await.mine_with_many_transactions(executions).await?
```
dinhani-cw, line 101: There is no need for miner here.
Miner will not be used because we are not going to generate new blocks in this flow.
We will import external blocks as-is, keeping the same numbers and transactions.

dinhani-cw (approved):
Would be nice to return Result instead of panicking in the cmp methods

dinhani-cw (approved):

Issue #204, opened 2024-02-13
Enhance Debugging for EVM Execution Failures


renancloudwalk:
# Problem
When an EVM transaction fails, the current setup does not provide enough context or information to pinpoint the cause of the failure. This makes debugging and resolving issues within EVM execution challenging and time-consuming.

# Proposed Solution
To improve the debugging process and reduce the time spent on issue resolution, the following enhancements are proposed:

1. **Detailed Logging:** Implement comprehensive logging at key points in the EVM execution path, including error states and exceptions. The logs should contain detailed information such as transaction inputs, account states, execution paths, and error messages.

2. **Error Context:** Ensure that all errors thrown during EVM execution include a context object that provides additional information about the state of the EVM at the time of the error.

3. **Diagnostic Tooling:** Develop or integrate existing diagnostic tools that can provide insights into the EVM state and transaction execution flow. These tools could include state inspectors, transaction tracers, and gas usage analyzers.

4. **Documentation:** Create documentation that maps out the EVM execution process within our system, highlighting common failure points and providing guidance on interpreting logs and errors.

# Action Items
- Audit the current EVM execution flow for opportunities to add or improve logging.
- Define a standard error object that includes relevant debugging information for EVM execution errors.
- Research and integrate diagnostic tools that can assist with debugging EVM execution.
- Produce documentation that assists developers in understanding the EVM execution process and debugging failures.

PR #205, opened 2024-02-13
Revm improve debugging

renancloudwalk:
https://github.com/cloudwalk/stratus/issues/204
- logs the state of the evm before and after an excecution
- reports when there is a failure to an account not being found

### DIFF
--- a/src/eth/storage/eth_storage.rs
+++ b/src/eth/storage/eth_storage.rs
@@ -82,11 +82,12 @@ pub trait EthStorage: Send + Sync {
             Some(account) => Ok(account),
             None => {
                 tracing::error!("Account not found: {}", address);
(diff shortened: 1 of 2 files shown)


src/eth/storage/eth_storage.rs
```diff
@@ -80,10 +80,14 @@ pub trait EthStorage: Send + Sync {
     async fn read_account(&self, address: &Address, point_in_time: &StoragePointInTime) -> anyhow::Result<Account> {
         match self.maybe_read_account(address, point_in_time).await? {
             Some(account) => Ok(account),
-            None => Ok(Account {
-                address: address.clone(),
-                ..Account::default()
-            }),
+            None => {
+                tracing::error!("Account not found: {}", address);
```
dinhani-cw, line 84: This is not an error.

PR #210, opened 2024-02-15
feat: pass gas and gas_price to evm when executing external transaction

dinhani-cw:
* When executing an external transaction, uses original gas_limit and gas_price to reproduce original behavior.

* When executing an eth_call or eth_sendRawTransaction, keep using max values (current behavior), but it can be changed to start using the input values if necessary instead of being hardcoded inside the REVM.

* Original gas_limit is calculated according to the transaction type.


renancloudwalk (commented):
👍

PR #211, opened 2024-02-15
Regression test for gas limit

### LINKED ISSUE #201: Fix and Regression Test Gas Limit
# Why?

The current implementation may incorrectly handle transactions with insufficient gas, leading to OutOfGas errors. Ensuring proper gas limit handling is crucial for transaction execution reliability.

# How?

Investigate and resolve the issue in the gas limit calculation or handling within the Revm execution process.
Implement a regression test that verifies transactions with various gas limits are handled correctly, ensuring no unexpected OutOfGas errors occur.

renancloudwalk:
[test transactions that exceed gas limit](https://github.com/cloudwalk/stratus/commit/b82138b9bd2d790e97f698314626895270acf056)

closes #201


e2e/test/helpers/account.ts
```diff
@@ -16,13 +16,13 @@ export class Account implements Addressable {
         return new Wallet(this.privateKey, ETHERJS);
     }
 
-    async signWeiTransfer(counterParty: string, amount: BigNumberish, nonce?: number): Promise<string> {
+    async signWeiTransfer(counterParty: string, amount: BigNumberish, gasLimit: BigNumberish, nonce?: number): Promise<string> {
```
dinhani-cw, line 19: Use a sensible default value for this new parameter to avoid the burden of all clients having to specify the gas when they don't care about it.

e2e/test/e2e-tx-broken-by-gas-limit.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    sendGetBalance,
+    sendRawTransaction,
+    sendReset,
+} from "./helpers/rpc";
+
+describe("GAS_LIMIT", () => {
+    var _tx: Transaction;
+    var _txHash: string;
+    var _block: Block;
+    var _txSentTimestamp: number;
```
dinhani-cw, line 27: Timestamp is not being use. You can remove it from this test.

e2e/test/e2e-tx-broken-by-gas-limit.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    ZERO,
+    fromHexTimestamp,
+    send,
+    sendGetBalance,
+    sendRawTransaction,
+    sendReset,
+} from "./helpers/rpc";
+
+describe("GAS_LIMIT", () => {
+    var _tx: Transaction;
```
dinhani-cw, line 24: Most of these values like `_txSentTimestamp` are not being used.
Please remove it.
They are declared here when we want to pass data from one step to another.
If this is not the case, they are not necessary here.

e2e/test/e2e-tx-broken-by-gas-limit.test.ts
```diff
@@ -0,0 +1,54 @@
...
+        let txSigned = await ALICE.signWeiTransfer(BOB.address, TEST_TRANSFER, 0, 0);
+        _txSentTimestamp = Math.floor(Date.now() / 1000);
+        _txHash = await sendRawTransaction(txSigned);
+        expect(_txHash).eq(undefined);
+    });
+    it("Block is created", async () => {
+        expect(await send("eth_blockNumber")).eq(ZERO);
+
+        _block = await send("eth_getBlockByNumber", [ONE, true]);
+        expect(_block).eq(null);
```
dinhani-cw, line 43: This does not match `block is created`. What is really being tested here?

e2e/test/e2e-tx-broken-by-gas-limit.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    var _tx: Transaction;
+    var _txHash: string;
+    var _block: Block;
+    var _txSentTimestamp: number;
+    var new_account: Account;
+
+    it("Resets blockchain", async () => {
+        await sendReset();
+    });
+    it("Send transaction", async () => {
```
dinhani-cw, line 33: Add a more descriptive name to this step like: `Sends transaction without enough gas`.

PR #212, opened 2024-02-15
chore: rename transaction input gas into gas_limit

renancloudwalk:


arthur-cw:
is this for compliance with Frontier? because I think `gas` is a valid field name

renancloudwalk:
@arthur-cw not only compliance but to make the naming more explicit

PR #219, opened 2024-02-16
feat: improving comparison in offline block importer

dinhani-cw:


src/eth/primitives/execution.rs
```diff
@@ -57,55 +56,70 @@ impl Execution {
...
         }
 
+        // compare logs length
         if self.logs.len() != receipt.logs.len() {
-            return Err(anyhow!(
-                "tx logs length mismatch, expected: {:?} got: {:?}",
-                receipt.logs.len(),
-                self.logs.len()
+            return log_and_err!(format!(
+                "logs length does not match | hash={} execution={} receipt={}",
```
carneiro-cw, line 79: I think here the correct would be "do not", since the subject is logs which is not 3rd-person singular. Same everywhere else where the subject is plural. https://testbook.com/key-differences/difference-between-do-and-does

carneiro-cw (approved):

PR #224, opened 2024-02-19
refactor: use UnixTime instead of u64 everywhere

dinhani-cw:
Codebase was inconsistent state where in some places Unix timestamps are represented as `u64`, and in other they are represents with the struct `UnixTime`.

Unify all representations using `UnixTime`.

### DIFF
--- a/static/schema/001-init.sql
+++ b/static/schema/001-init.sql
@@ -5,7 +5,7 @@ CREATE TABLE IF NOT EXISTS blocks (
     ,transactions_root BYTEA NOT NULL CHECK (LENGTH(transactions_root) = 32)
     ,gas NUMERIC NOT NULL CHECK (gas >= 0)
     ,logs_bloom BYTEA NOT NULL CHECK (LENGTH(logs_bloom) = 256)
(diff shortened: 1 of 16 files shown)


static/schema/001-init.sql
```diff
@@ -5,7 +5,7 @@ CREATE TABLE IF NOT EXISTS blocks (
     ,transactions_root BYTEA NOT NULL CHECK (LENGTH(transactions_root) = 32)
     ,gas NUMERIC NOT NULL CHECK (gas >= 0)
     ,logs_bloom BYTEA NOT NULL CHECK (LENGTH(logs_bloom) = 256)
-    ,timestamp_in_secs INTEGER NOT NULL CHECK (timestamp_in_secs >= 0) -- UNIQUE
+    ,timestamp_in_secs INTEGER NOT NULL CHECK (timestamp_in_secs >= 0) -- UNIQUE | TODO: INTEGER is the wrong type for timestamps
```
renancloudwalk, line 8: `bigint`

renancloudwalk (commented):
👍

PR #223, opened 2024-02-16
tests: pass all e2e-contract tests

carneiro-cw:
I had to patch some of the contracts that used functionality exclusive to hardhat. Also postgres still does not pass all the tests, ~but will when I finish this pr~ when this pr closes I'll start working on another one so postgres also passes the tests.

PR #229, opened 2024-02-19
chore: improve changes made in #223

carneiro-cw:
there are some improvements to be made for the changes in #223, namely a few changes on the new rpc method tests, feature flags, variable name changes, and improving evm_setNextBlockTimestamp implementation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,3 +102,13 @@ clone_on_ref_ptr = "warn"
 disallowed_names = "warn"
 manual_let_else = "warn"
 semicolon_if_nothing_returned = "warn"
+
+# ------------------------------------------------------------------------------
+# Features
+# ------------------------------------------------------------------------------
+
+[features]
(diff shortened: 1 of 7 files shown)


dinhani-cw:
At least for now I would prefer to just have two features: "development" and "production".

carneiro-cw:
> At least for now I would prefer to just have two features: "development" and "production". A one-to-one replacement of the Environment enum.

Are those features mutually exclusive? [The Cargo Book ](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features) states that "This should be avoided if at all possible, because it requires coordinating all uses of the package in the dependency graph to cooperate to avoid enabling them together.". Maybe we can have only a "development" feature, and have `not(feature = "development")` mean production,

dinhani-cw (approved):

dinhani-cw (approved):

PR #231, opened 2024-02-19
chore: show backtrace whenever there is an error

renancloudwalk:


carneiro-cw:
Just to add some context: 
This could be useful because later when an RpcError is converted to an ErrorObjectOwned we lose a lot of the information stored in the error, such as its trace. So in places where we use the `?` operator to deal with results in `rpc_server.rs` (eg. [here](https://github.com/cloudwalk/stratus/blob/9ae3dcdf93b21e9be9338416c7d565aa2d1efbac/src/eth/rpc/rpc_server.rs#L325)) we never log the trace.

One solution would be to match every result, and trace in the arm that matches errors. This obviously would make the code more verbose.

Maybe there is another place in the server pipeline further up where we can trace the RpcError before it gets converted to an ErrorObjectOwned.

dinhani-cw (approved):

PR #235, opened 2024-02-20
fix(e2e-contracts): wrap test return code in a conditional

mayconamaroCW:
Fixes the problem found by @carneiro-cw in which the script stops at the first successful test suite

we needed to make yield streamer public


carneiro-cw:
Can you pipe the stratus output to /dev/null so that we can see the runs?

dinhani-cw (approved):

PR #237, opened 2024-02-21
chore: add a loop for greater retries in order to gather current block

renancloudwalk:


renancloudwalk:
closed because we can solve most of the issues raising max_connections AND the acquiring pool config

PR #222, opened 2024-02-16
Move to new Storage traits

arthur-cw:
This PR splits the `EthStorage` trait into two new traits that represent Temporary and Permanent storages (for staging changes between blocks, and for committing the data to the DB, respectively).

I also added the NodeJS directories and package files to .gitignore

PR #238, opened 2024-02-21
feat: make importer work with current storage design

dinhani-cw:
**Main points**

This PR makes the block importer fully works with the in-memory `EthStorage` implementation, but performance is awful.

It works this way for each block to import:
* For each block to import, each transaction is re-executed and the temporary account changes are stored in the storage and the transaction execution changes are collected to a vector.
* After all transactions are re-executed in the block, the external block is converted to the internal block format with the related transactions and executions.
* The storage state is rolled-back using `reset` to its initial state before all temporary data was stored.
* The produced block with the is saved permanently to the storage.

The bottleneck of this flow is the `reset` method. Even the in-memory implementation is slow and can possibly be improved.

PostgreSQL is not working yet because `save_account_changes` method is not implemented, and it probably requires some changes to make it work. Saving temporary account changes to PostgreSQL without having a related block is not possible because the foreign key constraint that does not exist in the in-memory implementation, but exists in PostgreSQL.

It is clear that this approach is not the best and #222 is the correct way to handle this by splitting temporary and permanent storage.

**Additional points**

* I removed the `ConversionError` enum in favor of `anyhow` error because it is how errors are being handled in the entire application. There was no specific use that required `ConversionError` to exist.
* There are several possible conversion problems that I annotated with TODOs.

### DIFF
--- a/src/eth/primitives/log_mined.rs
+++ b/src/eth/primitives/log_mined.rs
@@ -59,6 +59,22 @@ impl LogMined {
 }
 
 // -----------------------------------------------------------------------------
+// Conversions: Other -> Self
+// -----------------------------------------------------------------------------
+impl From<EthersLog> for LogMined {
+    fn from(value: EthersLog) -> Self {
+        Self {
+            transaction_hash: value.transaction_hash.expect("log must have transaction_hash").into(),
+            transaction_index: value.transaction_index.expect("log must have transaction_index").into(),
+            log_index: value.log_index.expect("log must have log_index").into(),
+            block_number: value.block_number.expect("log must have block_number").into(),
+            block_hash: value.block_hash.expect("log must have block_hash").into(),
+            log: value.into(),
+        }
+    }
+}
+
+// -----------------------------------------------------------------------------
 // Conversions: Self -> Other
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


src/eth/primitives/log_mined.rs
```diff
@@ -58,6 +58,22 @@ impl LogMined {
     }
 }
 
+// -----------------------------------------------------------------------------
+// Conversions: Other -> Self
+// -----------------------------------------------------------------------------
+impl From<EthersLog> for LogMined {
+    fn from(value: EthersLog) -> Self {
+        Self {
```
carneiro-cw, line 66: Should we panic here? Why not impl TryFrom?

dinhani-cw:
@renancloudwalk No. I save only the account changes first. Then reset and save block, transactions and account changes again.

src/eth/primitives/log_mined.rs
dinhani-cw: Both approaches seems wrong. We need to ensure at the type level the input of this function have all fields populated, because they should have. They are logs that were added to a block, so these fields should never be None.

carneiro-cw (approved):

arthurmm-cloudwalk (approved):

PR #239, opened 2024-02-21
test: pin commit to repo clones

mayconamaroCW:
To avoid breaking Contracts E2E automated test, this PR checkouts each contract repo in a specific commit after cloning.

### DIFF
--- a/e2e-contracts/clone-contracts.sh
+++ b/e2e-contracts/clone-contracts.sh
@@ -11,6 +11,7 @@ source $(dirname $0)/_functions.sh
 # Clone a project to the projects directory.
 clone() {
     repo=$1
+    commit=$2
     target=repos/$repo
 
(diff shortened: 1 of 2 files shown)


e2e-contracts/clone-contracts.sh
```diff
@@ -33,9 +41,9 @@ clone() {
...
-clone compound-periphery
-clone brlc-yield-streamer
-clone brlc-pix-cashier
\ No newline at end of file
+clone brlc-multisig       918a226
+clone brlc-periphery      b8d507a
+clone brlc-token          55770b9
+clone compound-periphery  e4d68df
+clone brlc-yield-streamer e63d8ba
+clone brlc-pix-cashier    00cc007
```
renancloudwalk, line 49: so we need to track them manually then?

PR #252, opened 2024-02-22
chore: split inmemorystorage into permanent and temporary

### LINKED ISSUE #251: Split InMemory storage implementation into Temporary and Permanent
We need to split InMemoryStorage in two:

* `InMemoryStoragePermanent`: the same implementation, it is just a rename.
* `InMemoryStorageTemporary`: just a smaller scope of InMemoryStorage handling only account and slots changes, but without keeping track of historical values, block or transactions.

The reuse of the same storage struct for two different context (permanent and temporary) caused a issue that was very hard to debug, but easy to fix.

Splitting it in two can prevent this kind of problem of happening again because of mixed contexts.

carneiro-cw:
closes #251


arthur-cw (commented):
please add descriptive commentary and docs to describe why this is necessary

dinhani-cw (approved):

PR #256, opened 2024-02-22
chore: log read storage origin and track with metrics

### LINKED ISSUE #248: Track account and slot origin with logs and metrics
Both accounts and slots now can be retrieved from the temporary storage or permanent storage.

It will be very useful to have the following information in the related metrics and logs to help debug possible issues:
* found in the permanent storage
* found in the temporary storage
* not found, assuming default value

carneiro-cw:
closes #248


src/infra/metrics.rs
```diff
@@ -57,7 +57,7 @@ metrics! {
     histogram storage_check_conflicts{conflicted, success},
 
     "Duration of storage read_account operation."
-    histogram storage_read_account{point_in_time, success},
+    histogram storage_read_account{point_in_time, success, origin},
```
dinhani-cw, line 60: keep success always as the last parameter as all the other metrics

src/infra/metrics.rs
```diff
@@ -66,7 +66,7 @@ metrics! {
     histogram storage_read_logs{success},
 
     "Duration of storage read_slot operation."
-    histogram storage_read_slot{point_in_time, success},
+    histogram storage_read_slot{point_in_time, success, origin},
```
dinhani-cw, line 69: keep success always as the last parameter as all the other metrics
origin can be the first one because it is a category

src/infra/metrics.rs
```diff
@@ -147,6 +153,24 @@ fn into_labels(labels: Vec<(&'static str, LabelValue)>) -> Vec<MetricsLabel> {
         .collect()
 }
 
+/// Labels for storage types
+
+pub enum StorageType {
```
dinhani-cw, line 158: I don't think it makes sense to have this here or anywhere.
This module does not have any specifics enums from a specific context.

You can put in the in StratusStorage module, but I think just having a few constants there like "temp_storage", "perm_storage" and "default_value" should be enough.

carneiro-cw: Sure, I'll do that. 

As a side note: Shouldn't the constants be in this module either way? These labels don't concern the storage module, they are only relevant in the context of metrics. Sure this module does not have specific labels, but is there a reason that it shouldn't?

src/eth/storage/stratus_storage.rs
```diff
@@ -22,6 +22,10 @@ use crate::eth::storage::StorageError;
 use crate::eth::storage::TemporaryStorage;
 use crate::infra::metrics;
 
+static TEMP_LABEL: &str = "Temporary";
```
dinhani-cw, line 25: use lowercase labels, please

dinhani-cw (approved):

PR #261, opened 2024-02-22
chore: insert genesis and pre-funded account in 001-init.sql

### LINKED ISSUE #259: Insert accounts and genesis block directly into Postgres schema
For the Postgres storage implementation, it's required that every account be associated with a block number in which it was initialized. To solve that for pre-funded accounts, we can insert them and the genesis block itself directly in the schema.

carneiro-cw:
closes #259


arthur-cw:
pls add a comment section just to clarify why this is necessary

arthur-cw (approved):

PR #273, opened 2024-02-24
Add metrics to execution

renancloudwalk:


src/eth/executor.rs
```diff
@@ -103,6 +108,7 @@ impl EthExecutor {
 
                     // temporarily save state to next transactions from the same block
                     self.storage.save_account_changes_to_temp(execution.clone()).await?;
+                    metrics::inc_execution_and_commit(start.elapsed(), true);
```
dinhani-cw, line 111: It is not committing in the previous line. The commit is after the for loop finishes, so we should use a different name here.

PR #274, opened 2024-02-26
build: deny warnings when lint checking

mayconamaroCW:
### Changes

- `just lint-check` will now fail if there any warnings. Consequently, `Lint and Style checks` in GitHub will also fail.
- add flag to choose which version of nightly to use with the recipe. Example `just lint-check -2024-01-01`
- pin nightly version downloaded by `Lint and Style checks` at Jan 1, 2024.

Gives more visibility if a PR is introducing a warning


justfile
```diff
@@ -67,7 +67,7 @@ lint:
 
 # Stratus: Lint and check code formatting
 lint-check:
-    @just _lint --check
+    @just _lint --check "-D warnings -A unused-imports"
```
mayconamaroCW, line 70: redundant/unused imports are a recent feature, not sure if I should allow it or not

carneiro-cw (approved):

PR #283, opened 2024-02-26
perf: insert entire block at once (postgres)

carneiro-cw:
The upside to this change is that we can insert a block with all of its changes in one network call (instead of 9), the downside is that we get a huge query with 60 parameters.


e2e/test/e2e-json-rpc.test.ts
```diff
@@ -100,7 +100,7 @@ describe("JSON-RPC", () => {
             it("Should reset the changes when sending 0", async () => {
                 await send("evm_setNextBlockTimestamp", [0]);
                 await send("evm_mine", []);
-                expect((await latest()).timestamp).eq(Math.floor(Date.now() / 1000));
+                expect((await latest()).timestamp).approximately(Math.floor(Date.now() / 1000), 1);
```
mayconamaroCW, line 103: instead of a fixed delta, I think it is less error-prone to check if latest timestamp is no sooner than the timestamp of the function call, and no later than `now()` inside the check. We have an example of this in serial transfer -> block is created

mayconamaroCW (approved):

PR #287, opened 2024-02-27
chore: add workflow for performance tracking

renancloudwalk:


mayconamaroCW (changes requested):
Choose **one** of the following:
- Move `_setup-e2e2.yml` steps to `e2e-performance.yml` and erase the former. There is no need for a reusable workflow to exist just for one action to use it.
- Transform `runs-on` into a input with `ubuntu-latest` as a default. In this way, `e2e-performance` can use the reusable workflow `_setup-e2e.yml`

PR #296, opened 2024-02-29
feat: state validator binary

carneiro-cw:


src/eth/storage/stratus_storage.rs
```diff
@@ -220,4 +221,8 @@ impl StratusStorage {
             },
         }
     }
+
+    pub async fn get_slots_sample(&self, start: BlockNumber, end: BlockNumber, max_samples: u64, seed: u64) -> anyhow::Result<Vec<SlotSample>> {
```
dinhani-cw, line 225: All reader functions starts with `read_`.
Please follow the same pattern here.

src/eth/primitives/slot.rs
```diff
@@ -202,3 +204,11 @@ impl PgHasArrayType for SlotValue {
         <[u8; 32] as PgHasArrayType>::array_type_info()
     }
 }
+
+#[derive(Debug, sqlx::Decode)]
+pub struct SlotSample {
+    pub address: Address,
+    pub block_number: BlockNumber,
+    pub index: SlotIndex,
```
dinhani-cw, line 212: Can you use `Slot` directly here instead of `index` and `value`?

src/eth/storage/inmemory/inmemory_permanent.rs
```diff
@@ -353,6 +357,41 @@ impl PermanentStorage for InMemoryPermanentStorage {
...
         Ok(())
     }
+
+    async fn get_slots_sample(&self, start: BlockNumber, end: BlockNumber, max_samples: u64, seed: u64) -> anyhow::Result<Vec<SlotSample>> {
+        let state = self.lock_read().await;
+
+        let samples = state
+            .accounts
+            .iter()
+            .filter(|(_, account_info)| account_info.bytecode.get_current().is_some())
```
dinhani-cw, line 367: Create a function `is_contract` in the account struct to make this filter more clear in its intent.

src/eth/primitives/slot.rs
carneiro-cw: That what I attempted at first, but I couldn't make `query_file_as!` play nicely with nested structs. This is a known issue https://github.com/launchbadge/sqlx/discussions/1446. I could use the query function and then map the result. But I found adding the index and value as fields so that I could use `query_file_as!` nicer.

dinhani-cw (approved):

PR #300, opened 2024-03-01
chore: add the missing fields of a block to the header

carneiro-cw:
There are some fields that are constant, those are the uncle_hash, nonce, difficulty and total difficulty. This PR does not implement the calculation for the added fields when we create a new block, but it already works when importing blocks.


arthur-cw (commented):
`Difficulty` is totally useless for us, maybe we dont need to maintain all this code for it?

src/eth/primitives/block_nonce.rs
```diff
@@ -0,0 +1,82 @@
...
+use sqlx::encode::IsNull;
+use sqlx::error::BoxDynError;
+use sqlx::postgres::PgHasArrayType;
+use sqlx::Decode;
+
+use crate::gen_newtype_from;
+
+/// BlockNonce of an Ethereum account (wallet or contract).
+#[derive(Debug, Clone, Default, Eq, PartialEq, Hash, serde::Serialize, serde::Deserialize)]
+pub struct BlockNonce(H64);
```
arthur-cw, line 15: What is this for? 🤔

carneiro-cw: I copied the code from bytecode so i didn't have to reimplement everything (so the comment might make. But it is one of the fields returned by eth_getBlockByNumber (it should be constant afaik.

src/eth/primitives/block_header.rs
```diff
@@ -48,10 +63,21 @@ impl BlockHeader {
...
+            gas_limit: Gas::MAX,
             bloom: LogsBloom::default(),
             timestamp,
             parent_hash: number.prev().map(|n| n.hash()).unwrap_or(Hash::zero()),
+            author: Address::default(),
+            extra_data: Bytes::default(),
+            miner: Address::default(),
+            difficulty: Difficulty::default(),
+            receipts_root: Hash::zero(),
+            uncle_hash: Hash::zero(),
```
dinhani-cw, line 76: Empty hashes usually have special values like the ones defined here instead of being zero.

It is important for this to be correct because libraries like the Geth client in Go validates this information.

https://github.com/cloudwalk/stratus/blob/f3216247ee402489f53005951325fb865337635f/src/eth/primitives/block_header.rs#L27-L31

carneiro-cw: Roger

src/eth/primitives/block_header.rs
```diff
@@ -48,10 +63,21 @@ impl BlockHeader {
             number,
             hash: number.hash(),
             transactions_root: HASH_EMPTY_TRANSACTIONS_ROOT,
-            gas: Gas::ZERO,
+            gas_used: Gas::ZERO,
+            gas_limit: Gas::MAX,
```
dinhani-cw, line 67: Default to zero.

src/eth/primitives/block_nonce.rs
arthur-cw: how is it different from the `Nonce` type?

dinhani-cw: Just use the  `Nonce` type, they represent the same thing.

arthur-cw:
`Difficulty` and possibly other values are internal params and can't be changed by RPC calls

carneiro-cw:
> `Difficulty` and possibly other values are internal params and can't be changed by RPC calls, that's why I don't think we need to maintain them

Since ethereum became PoS the fields difficulty, totalDifficulty, nonce and uncleHash became constant. I'd agree that we can probably safely remove them.

src/eth/primitives/block_nonce.rs
carneiro-cw: They are not represented using the same data types though, a nonce is an U256, while the BlockNonce is a H64.

carneiro-cw: So unlike an account nonce which is a number with 256 bits, the block nonce is a hash with 64 bits.

arthur-cw: oh now I got this, this Nonce is the mining algorithm nonce. Ok, we can have another type, but BlockNonce is not so good of a name. I suggest `MinerNonce` or `MiningNonce`.

But I'm thinking maybe we can have more general types to cover these kind of situations, like `Hashable` or something instead of a bunch of wrappers with the same types

arthur-cw:
>Since ethereum became PoS the fields difficulty, totalDifficulty, nonce and uncleHash became constant. I'd agree that we can probably safely remove them.

👍

src/eth/primitives/block_nonce.rs
carneiro-cw: I renamed it to MinerNonce for now. But I agree that having more general types would be good, even for the other eth primitives. So instead of having `struct Gas(ethereum_types::U256)` we could have `struct Gas(stratus::U256)` and implement traits such as sqlx conversions and whatever else is specific to `U256` and not `Gas` directly on `stratus::U256`.

carneiro-cw:
@dinhani-cw 
> Are you also considering the external blocks we are importing?

Yes, that was my primary consideration. I have not confirmed that Frontier won't ever give a different value for one of those fields, but I have confirmed that for the first 500k blocks of the mainnet those fields are constant.

PR #302, opened 2024-03-04
chore: just put the error plainly on the return

renancloudwalk:
Can't see any reason to obfuscate the error on an error log


src/eth/evm/revm.rs
```diff
@@ -103,10 +103,7 @@ impl Evm for Revm {
 
         let result = match evm_result {
             Ok(result) => Ok(parse_revm_execution(result, session.input, session.storage_changes)),
-            Err(e) => {
-                tracing::warn!(reason = ?e, "evm execution error");
-                Err(anyhow!("Error executing EVM transaction. Check logs for more information."))
-            }
+            Err(e) => Err(anyhow!("Error executing EVM transaction. {:?}", e)),
```
mayconamaroCW, line 106: does it make sense to have the error both on `Err` return and the `tracing::warn!` ?
Better to have more than not enough 😬

PR #313, opened 2024-03-05
chore: use u64 to represent most integer newtypes in rust, and numeric in pgsql

carneiro-cw:


renancloudwalk (commented):
Later we will need to evaluate the performance implications of block_numbers as numeric

renancloudwalk (approved):

carneiro-cw:
Have we? We discussed that there might be performance hits using and indexed numeric value, but I thought we were going with everything as numeric at first.

mayconamaroCW (dismissed):

PR #314, opened 2024-03-05
feat: run block importer with postgres instead of inmemory

dinhani-cw:
* Test is running with PostgreSQL.
* Testcontainer initial adoption.

### DIFF
--- a/tests/test_import_offline_snapshot.rs
+++ b/tests/test_import_offline_snapshot.rs
@@ -1,26 +1,35 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
+use itertools::Itertools;
 use stratus::config::CommonConfig;
 use stratus::eth::primitives::ExternalBlock;
 use stratus::eth::primitives::ExternalReceipt;
-use stratus::eth::storage::InMemoryPermanentStorage;
+use stratus::eth::primitives::StoragePointInTime;
 use stratus::eth::storage::InMemoryPermanentStorageState;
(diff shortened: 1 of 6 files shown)


tests/test_import_offline_snapshot.rs
```diff
@@ -60,3 +76,39 @@ async fn test_import_offline_snapshot() {
...
+    // init docker container (this should be extract to a reusable module if going to be used in other tests)
+    let pg_image = RunnableImage::from(PostgresImage::default().with_user("postgres").with_password("123").with_db_name("stratus"))
+        .with_mapped_port((5432, 5432))
+        .with_volume(("./static/schema/001-init.sql", "/docker-entrypoint-initdb.d/001-schema.sql"))
+        .with_volume(("./static/schema/002-schema-external-rpc.sql", "/docker-entrypoint-initdb.d/002-schema.sql"))
+        .with_tag("16.1");
+    let pg_container = docker.run(pg_image);
+
+    // init postgres client
+    let pg = Postgres::new("postgres://postgres:123@localhost:5432/stratus").await.unwrap();
```
renancloudwalk, line 90: use clap?

mayconamaroCW (approved):

tests/test_import_offline_snapshot.rs
dinhani-cw: No need for clap here, unless we want to connect to some external database.

The connection URL is hardcoded here because it is using the same parameters defined in the test-container, so assuming this test only wants to use the test container, the URL wil never be different.

PR #317, opened 2024-03-06
Analysed optimization

renancloudwalk:
WIP

### DIFF
--- a/src/eth/evm/revm.rs
+++ b/src/eth/evm/revm.rs
@@ -9,13 +9,12 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Instant;
 
-use revm::primitives::KECCAK_EMPTY;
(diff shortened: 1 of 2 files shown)


src/eth/evm/revm.rs
```diff
@@ -174,7 +182,18 @@ impl Database for RevmDatabaseSession {
                 .insert(account.address.clone(), ExecutionAccountChanges::from_existing_account(account.clone()));
         }
 
-        Ok(Some(account.into()))
+        let bytecode = if let Some(bytes) = account.bytecode {
+            Some(self.bytecodes.entry(address).or_insert(to_analysed(bytes.into())).clone())
```
dinhani-cw, line 186: You cannot cache the bytecode until you are sure the transaction is confirmed, otherwise you can have an invalid bytecode associated with an address.

This is the reason I populated the cache only after the transaction is finished and confirmed to be a success.

PR #336, opened 2024-03-11
remove schema and update justfile

arthur-cw:
retrieve from capybara-migrations instead
still lacking credentials cc @lmarangoni


carneiro-cw:
I think this schema does not insert the special account with max balance

renancloudwalk (changes requested):
needs to delete everything under the schema, there are 3 files there.
Also, needs to update the docker-compose commands in two places

arthurmm-cloudwalk:
@arthur-cw is there any reason we can't let the schema in the repository ?

arthurmm-cloudwalk (changes requested):
I don't see a reason for the schema to be outside the repository. Is there any reason ?

renancloudwalk:
> I don't see a reason for the schema to be outside the repository. Is there any reason ?

It will conflict with capybara-migrations schema

Issue #329, opened 2024-03-08
Compute and store code_hash


dinhani-cw:
[`code_hash`](https://bluealloy.github.io/revm/docs/revm/interpreter/primitives/struct.AccountInfo.html#structfield.code_hash) is an attribute of an EVM account, being used both in REVM and Frontier.

Even if it is not required, computing it and storing it in the storage can be bring benefits to the application.

As example, REVM will call [`code_by_hash`](https://bluealloy.github.io/revm/docs/revm/trait.Database.html#tymethod.code_by_hash) to get the bytecode by its hash when the bytecode is not present in the account. This allows us to return a pre-analysed bytecode cached somewhere.

It also avoid recalculating the bytecode hash everytime it is returned to check if the analysed version is already cached. It will be calculated only once the account is created.

PR #339, opened 2024-03-12
Add code_hash

arthur-cw:
Defines and save the digest of an account's bytecode (if bytecode is null, digest is hash of empty string, constant value)

closes https://github.com/cloudwalk/stratus/issues/329


arthurmm-cloudwalk (dismissed):

src/eth/storage/inmemory/inmemory_permanent.rs
```diff
@@ -479,11 +480,15 @@ impl InMemoryPermanentAccount {
 
     /// Converts itself to an account at a point-in-time.
     pub fn to_account(&self, point_in_time: &StoragePointInTime) -> Account {
+        let bytecode = self.bytecode.get_at_point(point_in_time).unwrap_or_default();
+        let code_hash = CodeHash::from_bytecode(bytecode.clone());
```
dinhani-cw, line 484: It seems we are calculating the hash every time the conversion happens.

We should persist the code hash to the storage once and just read it when retrieving information

PR #342, opened 2024-03-12
chore: attempt to optimize tokio scheduler

renancloudwalk:


carneiro-cw:
Another option to look into is [max_lifetime](https://docs.rs/sqlx/latest/sqlx/pool/struct.PoolOptions.html#method.max_lifetime). Do note that "Infinite connections are not recommended due to the unfortunate reality of memory/resource leaks on the database-side. It is better to retire connections periodically (even if only once daily) to allow the database the opportunity to clean up data structures (parse trees, query metadata caches, thread-local storage, etc.) that are associated with a session."

arthur-cw:
We can try to specify the runtime params more granularly, for example the number of available cores.

PR #387, opened 2024-03-18
feat: adds historical_balances to csv exporter

brunoantonieto-cw:
Adds `historical_balances` table to CsvExporter


src/eth/primitives/execution_value_change.rs
```diff
@@ -15,6 +19,15 @@ pub enum ValueState<T> {
     NotSet,
 }
 
+impl Display for ExecutionValueChange<wei::Wei> {
```
dinhani-cw, line 22: No need for this Display implementation.

dinhani-cw (approved):

PR #394, opened 2024-03-19
feat: add historical slots table to CsvExporter

brunoantonieto-cw:
adds `historical_slots` table to CsvExporter


dinhani-cw (changes requested):
Adjust headers

dinhani-cw (approved):

PR #428, opened 2024-03-22
chore: add rocksdb

renancloudwalk:


mayconamaroCW (commented):
.

mayconamaroCW (changes requested):
I need to investigate why the cache didn't improve performance of CI, please don't merge

PR #444, opened 2024-03-25
fix: csv exporter bytearray serialization

### LINKED ISSUE #429: CSV exporter: Handle missing column `code_hash`
`code_hash` was introduced by #339 but that wasn't updated in the CSV exporter.

### LINKED ISSUE #430: CSV exporter: Fix fields of type `bytea` being written in incorrect format
Values for these fields are different than the same when loaded from Postgres directly.

marcospb19-cw:
Fixes #430.
Fixes #429.


src/eth/storage/csv/csv_exporter.rs
```diff
@@ -516,11 +518,22 @@ fn csv_writer(base_path: &'static str, number: BlockNumber, headers: &[&'static
...
 }
 
+/// Convert a byte sequence to a `bytea` representation that is parseable by Postgres.
+fn to_bytea(bytes: impl AsRef<[u8]>) -> String {
+    let hex = hex::encode(bytes.as_ref());
+    format!("\\x{hex}")
+}
+
+fn u256_to_bytea(integer: U256) -> String {
+    let bytes: Vec<u8> = integer.as_ref().iter().copied().rev().flat_map(u64::to_be_bytes).collect();
```
marcospb19-cw, line 533: I need more eyes on this `u64::to_be_bytes`.

marcospb19-cw: We tested that it produces the expected result in my local machine, but could it change for other machines? (does it even matter?)

mayconamaroCW: we already use `to_be_bytes` in other parts of the code ([example](https://github.com/cloudwalk/stratus/blob/b8c81b22a4a7ad6b224e35ba67fd48d2cddba931/src/eth/primitives/block_number.rs#L166)). to this day it never was a problem.

dinhani-cw (approved):

brunoantonieto-cw (approved):

PR #1035, opened 2024-06-07
chore: reorder dockerfile for caching

mayconamaroCW:


mayconamaroCW:
Only worked for local building, our github actions builder did not use the cache. I'll retry with another approach later.

PR #1067, opened 2024-06-11
test: add linters to E2E tests

EvgeniiZaitsevCW:
1. The [ESLint](https://eslint.org/) and [Solhint](https://github.com/protofire/solhint) linters have been added and configured in folder `e2e` to check and auto-format TypeScript and JavaScript files.
2. The `ESLint` tool has been extended with the [ESLint Stylistic plugin](https://eslint.style/) to check and correct formatting issues (indents, spaces, etc.).
3. The existing [Prettier tool](https://prettier.io/) has been extended to auto-format the Solidity code.
4. The `.prettierignore` file has been added to make `Prettier` ignore auto-generated folders (`artifacts`, `cache`, `typechain-types`).
5. The errors and warnings found by the new linters have been fixed in the TypeScript, JavaScript and Solidity files.
6. The `e2e-lint` option in `justfile` has been corrected to run the new linters instead of Prettier. 
7. The option to run Prettier has been renamed to `e2e-prettier` in `justfile`.
8. Some unused code related to changing of the RPC provider has been removed from the `e2e/test/helpers/rpc.ts` file. That code is currently used in the `e2e-contracts` folder.
9. IMPORTANT! `Prettier` not always gives the good results of code formatting. So if you run `just e2e-prettier` you'll still find some possible changes in the code, but they are not applied intentionally due to the existing version seems more readable than that Prettier provides. It is also the reason why Prettier has not been used for check the formatting (e.g. for Solidity files through [solhint-plugin-prettier](https://github.com/fvictorio/solhint-plugin-prettier)). Linters are more suitable here. Unfortunately, no good supported formatting linter for Solidity was found. Solhint supported formatting checks up to version 3, but then dropped that (see this release note [here](https://github.com/protofire/solhint/releases/tag/v3.0.0)). So it is expected that `Prettier` will be used to  auto-format Solidity code, but then the results will be checked and inappropriate changes will be rolled back.
10. An example of a well-known issue with `Prettier` when you apply it is applied to Solidity is formatting intentional multi-line entities into a single line. E.g. the following code:
    ```solidity
    event Add(
        address indexed account,
        uint256 amount
    );
    ```
    will be changed as:
    ```solidity
    event Add(address indexed account, uint256 amount);
    ```
    A workaround to prevent that is use a comment on the first line, like:
    ```solidity
    event Add(
        address indexed account, //
        uint256 amount
    );
    ```
    But it looks a bit awkward.

### DIFF
--- a/e2e/.prettierignore
+++ b/e2e/.prettierignore
@@ -1,0 +1,4 @@
+# Files and folders that Prettier must ignore
+artifacts/**
+cache/**
+typechain-types/**
--- a/e2e/.prettierrc.js
+++ b/e2e/.prettierrc.js
@@ -1,5 +1,5 @@
 module.exports = {
-    plugins: ["@trivago/prettier-plugin-sort-imports"],
+    plugins: ["@trivago/prettier-plugin-sort-imports", "prettier-plugin-solidity"],
     // lines
     printWidth: 120,
     tabWidth: 4,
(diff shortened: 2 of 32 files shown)


EvgeniiZaitsevCW:
Closed as have too many issues on Codacy and I can't resolve them in the near future

PR #1117, opened 2024-06-14
feat: remove async from PermanentStorage and TemporaryStorage

dinhani-cw:
Current implementation are sync, but the trait lies saying they are async.
So lets make it more clear and handle it better.


marcospb19-cw (approved):
:)

PR #1224, opened 2024-06-24
fix: logs queries after last mined block should return nothing

dinhani-cw:


e2e/cloudwalk-contracts/integration/test/relayer.test.ts
```diff
@@ -78,14 +78,12 @@ describe("Relayer integration test", function () {
...
                     try {
-                        const tx = await brlcToken
-                            .connect(sender)
-                            .transfer(receiver.address, amount, {
-                                gasPrice: 0,
-                                gasLimit: GAS_LIMIT_OVERRIDE,
-                                type: 0,
-                                nonce: nonces[senderIndex],
-                            });
+                        const tx = await brlcToken.connect(sender).transfer(receiver.address, amount, {
```
dinhani-cw, line 81: Just linter/formatter here.

PR #1233, opened 2024-06-25
add consensus metrics

gabriel-aranha-cw:


renancloudwalk (dismissed):
I think it's important for us to track the readiness checks and tell how much time it spends ready and not-ready

renancloudwalk (approved):

Issue #1270, opened 2024-06-28
Fix transaction execution save when empty


gabriel-aranha-cw:
Fix empty transaction execution save log. Currently it is not saving if empty, but it should.
https://github.com/cloudwalk/stratus/blob/main/src/eth/consensus/mod.rs#L415

PR #1271, opened 2024-06-28
Fix empty transaction execution save log

gabriel-aranha-cw:
Fix empty transaction execution save log. Currently it is not saving if empty, but it should.


gabriel-aranha-cw:
https://github.com/cloudwalk/stratus/issues/1270

Issue #1221, opened 2024-06-24
Implement Block Appending Logic in Raft Consensus Module


renancloudwalk:
# Why:
To ensure the consistency of log entries in the Raft consensus module, it is necessary to validate the prevLogIndex before appending a new block. This validation ensures that logs are appended in the correct order and helps prevent inconsistencies that could arise from mismatched log indices.

# How:

- [ ] Modify the append_block_commit function to include a check for prevLogIndex.
- [ ] Update the log appending logic to handle cases where prevLogIndex matches the current index by appending the new block.
- [ ] Implement error handling for cases where prevLogIndex does not match, possibly rejecting the append or resolving log inconsistencies.
- [ ] Add unit tests to cover different scenarios of block appending, ensuring that the logic works as expected in both matching and non-matching prevLogIndex cases.

carneiro-cw:
We've shelved raft for now, closing.

PR #1272, opened 2024-06-28
enha: replace block number with log index

gabriel-aranha-cw:
This PR implements the log index as specified by RAFT, replacing the placeholder block number.


gabriel-aranha-cw:
closes https://github.com/cloudwalk/stratus/issues/1221

renancloudwalk (changes requested):
please move storage back inside Consensus, we will use it to commit the data

chaos/experiments/leader-election.sh
```diff
@@ -132,7 +132,7 @@ check_leader() {
     # Check the response for specific strings to determine the node status
     if [[ "$response" == *"append_transaction_executions called on leader node"* ]]; then
         return 0 # Success exit code for leader
-    elif [[ "$response" == *"APPEND_SUCCESS"* ]]; then
```
renancloudwalk, line 135: I think this should remain OR use a specific message, we actually could change this function to probe several orther behaviors

renancloudwalk (approved):
nice job!

chaos/experiments/leader-election.sh
gabriel-aranha-cw: I will have to improve these checks when I work on protobuf definitions soon. I will leave it to the next PR

PR #1298, opened 2024-06-30
Propagate executions into memory

renancloudwalk:


src/eth/storage/inmemory/inmemory_temporary.rs
```diff
@@ -190,6 +190,14 @@ impl TemporaryStorage for InMemoryTemporaryStorage {
         Ok(())
     }
 
+    // appends transactions to the pending block
+    // also make them available as a cache for queries
+    fn append_transaction(&self, tx: TransactionExecution) -> anyhow::Result<()> {
```
dinhani-cw, line 195: Why not use `save_execution` instead? It already updates all necessary internal information.

Calling this function will add the transaction, but the other information like pending nonce, balance, etc., will be outdated.

renancloudwalk: it needs to be replicated exactly as is.
save_execution checks for conflicts, which would lead to changes...

renancloudwalk: Another thing, the execution that came from propagation has no `#changes` which would break save_transaction

PR #1315, opened 2024-07-02
feat: tracing performance

dinhani-cw:


marcospb19-cw (approved):
Much better than my solution, LGTM.

PR #1319, opened 2024-07-02
Fill block `size` field

### LINKED ISSUE #1289: Block size field is null
Block size field is marked as required by some TypeScript libs that are breaking because we are returning null.

We need to compute and populate it.

marcospb19-cw:
Fixes #1289.


mayconamaroCW (approved):
### Test report

Tested locally with tx-replayer. It passes now ✅

PR #1359, opened 2024-07-04
rpc: log cleaned subscriptions

marcospb19-cw:


mayconamaroCW (dismissed):

src/eth/rpc/rpc_subscriptions.rs
```diff
@@ -77,14 +78,53 @@ impl RpcSubscriptions {
...
-                    inner_map.retain(|_, s| not(s.sink.is_closed()));
-                    not(inner_map.is_empty())
+                subs.pending_txs.write().await.retain(|_, sub| {
+                    let should_keep = not(sub.sink.is_closed());
+                    if !should_keep {
+                        pending_txs_subs_cleaned.push(sub.client.clone());
+                    }
+                    should_keep
+                });
+                subs.new_heads.write().await.retain(|_, sub| {
```
dinhani-cw, line 94: Consider using `partition_map` from `itertools`.
It will return two Vecs, one with the removed subscriptions and the other with the ones to keep.

marcospb19-cw: Cool! Didn't know about this one, only the 'partition' in std::Iterator.

I'll try it out.

marcospb19-cw: Tried it and found that the code becomes too verbose and less readable, so I'll merge without this, if you wanna take a shot too tho, I'd say go for it.

(I need to mark as resolved to proceed with merging.)

PR #1360, opened 2024-07-05
storage: perform checks on block numbers

marcospb19-cw:


src/eth/storage/stratus_storage.rs
```diff
@@ -52,7 +60,11 @@ impl StratusStorage {
 
     /// Creates a new storage with the specified temporary and permanent implementations.
     pub fn new(temp: Box<dyn TemporaryStorage>, perm: Box<dyn PermanentStorage>) -> Self {
-        Self { temp, perm }
+        Self {
+            temp,
+            perm,
+            last_saved_block_number: AtomicU64::new(0),
```
dinhani-cw, line 66: I think this attribute is not necessary.

We already have the necessary values being set in `self.perm.read_mined_block_number()` and `self.temp.read_pending_block_number()`.

We should check the invariants based on these values.

src/eth/storage/stratus_storage.rs
```diff
@@ -337,6 +338,31 @@ impl StratusStorage {
...
+                ?new_block_number,
+                ?last_mined_block_number,
+                gap_size,
+                "new block number isn't expected based on previous one, some might've been skipped or order is broken",
+            );
+        }
+
+        // check temporary storage block number, if set
+        if let Some(pending_block_number) = self.temp.read_pending_block_number()? {
+            if new_block_number != pending_block_number {
```
dinhani-cw, line 357: This check will always be true.
The block being saved is already closed and there is a new pending block being mined.
Probably you want to check that the `new_block_number == (pending_block_number - 1)`.

Example:
Block 1 is closed and being saved. Block 2 is the pending one.

marcospb19-cw: oh yeah I did a local test and already fixed it, ty

PR #1361, opened 2024-07-05
perf: remove nonce depencency, split graph components for parallel relaying

carneiro-cw:
Adding a nonce dependency obviously defeated the whole purpose of resigning the transactions. But it fixed a logic error this also fixes the same issue but does so without compromising on performance.

### DIFF
--- a/src/eth/relayer/external.rs
+++ b/src/eth/relayer/external.rs
@@ -532,14 +532,9 @@ impl ExternalRelayer {
         }
     }
 
-    /// Relays a dag by removing its roots and sending them consecutively. Returns `Ok` if we confirmed that all transactions
-    /// had the same receipts, returns `Err` if one or more transactions had receipts mismatches. The mismatches are saved
-    /// on the `mismatches` table in pgsql, or in ./data as a fallback.
-    #[tracing::instrument(name = "external_relayer::relay_dag", skip_all)]
-    async fn relay_dag(&self, mut dag: TransactionDag) -> MismatchedBlocks {
+    /// Relays a dag by removing its roots and sending them consecutively.
+    async fn relay_component(&self, mut dag: TransactionDag) -> MismatchedBlocks {
         let start = Instant::now();
-
-        tracing::debug!("relaying transactions");
 
         let mut results = vec![];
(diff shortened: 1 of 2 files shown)


src/eth/relayer/transaction_dag.rs
```diff
@@ -128,18 +128,48 @@ impl TransactionDag {
...
+        while dag.node_count() > 0 {
+            let mut root = None;
+
+            // find a root
+            for index in dag.node_identifiers() {
+                if dag.neighbors_directed(index, petgraph::Direction::Incoming).next().is_none() {
+                    root = Some(index);
+                    break;
                 }
             }
```
marcospb19-cw, line 144: lil suggestion for no mutability
```rust
let root = dag
    .node_identifiers()
    .position(|index| dag.neighbors_directed(index, Incoming).next().is_none());
```

---

instead of running this search multiple times inside of this `while dag.node_count() > 0 {` loop, I wonder if you can't just run it once, grab all roots, and then calculate the value of `ret` for each one of them.

src/eth/relayer/external.rs
```diff
@@ -562,6 +557,21 @@ impl ExternalRelayer {
...
+
+    // Split the dag and relay its components.
+    #[tracing::instrument(name = "external_relayer::relay_dag", skip_all)]
+    async fn relay_dag(&self, dag: TransactionDag) -> MismatchedBlocks {
+        let start = Instant::now();
+
+        tracing::debug!("relaying transactions");
+
+        let futures = dag.split_components().into_iter().map(|dag| self.relay_component(dag));
+        let results = join_all(futures).await;
```
marcospb19-cw, line 571: It shooooouldn't be a problem, but I'd use `buffered` on this one too instead of the `join_all` call, probably with a large buffer size tho.

carneiro-cw: True, will improve this in the next PR.

src/eth/relayer/transaction_dag.rs
carneiro-cw: I could run it only once, will improve in the next PR.

PR #1363, opened 2024-07-05
fix: gracefully shutdown in consensus code

marcospb19-cw:


dinhani-cw (dismissed):

src/eth/consensus/mod.rs
```diff
@@ -750,8 +779,14 @@ impl Consensus {
 
     /// Handles the propagation of log entries to peers in the consensus network.
     async fn handle_peer_propagation(mut peer: Peer, consensus: Arc<Consensus>) {
+        const TASK_NAME: &str = "consensus::propagate";
+
         let mut log_entry_queue: Vec<LogEntryData> = Vec::new();
         loop {
+            if GlobalState::warn_if_shutdown(TASK_NAME) {
+                return;
+            };
```
marcospb19-cw, line 788: Added this in the second commit, but I'm still confused about what should I do in the await points below this piece of code.

```rust
let mut receiver_lock = peer.receiver.lock().await;
match receiver_lock.recv().await {
    Ok(log_entry) => {
        log_entry_queue.push(log_entry);
    }
    Err(e) => {
        tracing::warn!("Error receiving log entry for peer {:?}: {:?}", peer.client, e);
    }
}
drop(receiver_lock);
```

Ideally, I'd put that whole block in an async block and use it as a branch in `select!`, however I'm a little afraid of messing with that piece, it seem susceptible to deadlocks 🤔.

I'll ask @renancloudwalk about it, since he is the one who wrote it and might be more confident modifying.

renancloudwalk: if it shuts down, it shuts down, this shouldnt be a cause for a deadlock

renancloudwalk (dismissed):

PR #1366, opened 2024-07-05
feat: consensus regressive sync

gabriel-aranha-cw:
This PR adds append entry logs replication to followers in accordance with RAFT specification.


renancloudwalk (commented):
looks good so far

chaos/experiments/leader-election.sh
```diff
@@ -275,9 +275,7 @@ run_test() {
             fi
         done
 
-        if [ $num_instances -eq 2 ]; then
-            sleep 40 # wait for leader election before raising the other instance to avoid split vote
-        fi
+        sleep 40 # wait for leader election before raising the other instance to avoid split vote
```
renancloudwalk, line 278: why not, if gt 1?

renancloudwalk (approved):

chaos/experiments/leader-election.sh
gabriel-aranha-cw: Done, thanks

PR #1394, opened 2024-07-10
feat: request replication

carneiro-cw:


src/eth/evm/revm.rs
```diff
@@ -111,7 +111,16 @@ impl Evm for Revm {
...
         // configure session
         let evm = &mut self.evm;
-        evm.db_mut().reset(input.clone());
+        let db = evm.db_mut();
+        db.reset(input.clone());
+
+        // if this stratus instance is the receiver of request replications we circumvent possible
+        // nonce mismatches by setting it to the latest nonce of the sender.
+        let nonce = if cfg!(feature = "request-replication-test-receiver") {
+            db.basic(input.from.into())?.map(|acc| acc.nonce)
```
dinhani-cw, line 120: I think you can set the Nonce to None instead of loading it.

eth_call already does that.

carneiro-cw: Makes sense, will quickly test this and change it if it works.

carneiro-cw: You're correct, as per the doc as well:

> Caution: If set to None, then nonce validation against the account's nonce is skipped: [InvalidTransaction::NonceTooHigh][InvalidTransaction::NonceTooLow]

PR #1449, opened 2024-07-15
fix: storage shutdown when receiving `SIGTERM`

marcospb19-cw:
it wasn't finishing properly cause other threads were blocked dropping
our Rocks storage, in turn, the `main` thread was unblocked and
exited early

---

After writing this, it seems like a more general solution is to wait for all threads and tasks to finish.

However, it's still possible to shoot you on the foot by send an `Arc` clone to a spawned thread or task and forgetting about it.

This one, although ugly, seems to work and make it hard to shoot your foot in the process.

An theoretical problem is that RocksDB lost the potential to shut down in parallel with other things, but I can't think of a simple way to do it while guaranteeing proper shutdown.

### DIFF
--- a/src/bin/importer_offline.rs
+++ b/src/bin/importer_offline.rs
@@ -93,6 +93,9 @@ async fn run(config: ImporterOfflineConfig) -> anyhow::Result<()> {
         .join()
         .expect("'block-importer' thread panic'ed instead of properly returning an error");
 
+    // Explicitly block the `main` thread to drop the storage.
+    drop(storage);
+
     Ok(())
 }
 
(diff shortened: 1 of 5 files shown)


marcospb19-cw:
After writing this, it seems like a more general solution is to wait for all threads and tasks to finish.

However, it's still possible to shoot you on the foot by send an `Arc` clone to a spawned thread or task and forgetting about it.

This one, although ugly, seems to work and make it hard to shoot your foot in the process.

An theoretical problem is that RocksDB lost the potential to shut down in parallel with other things, but I can't think of a simple way to do it while guaranteeing proper shutdown.

PR #1463, opened 2024-07-16
enha: create max subscription limit per client

gabriel-aranha-cw:
This PR implements a configurable limit to the max number of subscriptions from `eth_subscribe`, using a default of 15.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -263,7 +263,7 @@ pub struct StratusConfig {
     pub max_connections: u32,
 
     /// JSON-RPC max active subscriptions per client.
-    #[arg(long = "max-subscriptions", env = "MAX_SUBSCRIPTIONS", default_value = "10")]
+    #[arg(long = "max-subscriptions", env = "MAX_SUBSCRIPTIONS", default_value = "15")]
     pub max_subscriptions: u32,
 
     #[clap(flatten)]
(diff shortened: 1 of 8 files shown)


src/config.rs
```diff
@@ -262,6 +262,10 @@ pub struct StratusConfig {
     #[arg(long = "max-connections", env = "MAX_CONNECTIONS", default_value = "200")]
     pub max_connections: u32,
 
+    /// JSON-RPC max active subscriptions per client.
+    #[arg(long = "max-subscriptions", env = "MAX_SUBSCRIPTIONS", default_value = "10")]
```
dinhani-cw, line 266: Here the default value still 10.

In another PR we should extract this to a separete struct RpcServerConfig.

gabriel-aranha-cw: Mismatch fixed. Will take this refactoring in a next PR.

src/eth/rpc/rpc_subscriptions.rs
```diff
@@ -300,6 +301,29 @@ pub struct RpcSubscriptionsConnected {
...
+            .read()
+            .await
+            .values()
+            .flat_map(HashMap::values)
+            .filter(|s| s.client == *client)
+            .count();
+        tracing::info!(%pending_txs, %new_heads, %logs, "current client subscriptions");
+
+        if pending_txs + new_heads + logs >= max_subscriptions as usize {
+            return Err(RpcError::SubscriptionInvalid {
```
dinhani-cw, line 319: You need to create a new entry in the error enum for this validation.
The error being used here will give the wrong reason for the client.

PR #1471, opened 2024-07-17
ci: add features input in build binary

mayconamaroCW:


mayconamaroCW:
Codacy was bypassed because it was complaining of a problem that was fixed, and it was suggesting to do exactly what was already in the code. The semgrep linter seems to bug in multiline commands.

PR #1475, opened 2024-07-17
chore: add check based on latest imported block

renancloudwalk:


src/eth/consensus/mod.rs
```diff
@@ -607,6 +607,30 @@ impl Consensus {
     }
 
     pub async fn should_serve(&self) -> bool {
+        if self.importer_config.clone().is_some() {
```
carneiro-cw, line 610: I don't think this clone is needed

carneiro-cw (dismissed):

src/eth/rpc/rpc_server.rs
```diff
@@ -380,7 +360,7 @@ async fn stratus_get_subscriptions(_: Params<'_>, ctx: Arc<RpcContext>, ext: Ext
 // -----------------------------------------------------------------------------
 
 async fn net_listening(params: Params<'_>, arc: Arc<RpcContext>, ext: Extensions) -> Result<JsonValue, RpcError> {
-    stratus_readiness(params, arc, ext).await
+    stratus_health(params, arc, ext).await
```
dinhani-cw, line 363: If you are going to put this check inside `net_listening`, we will need some tracing here because connected clients can also use this operation, not only the health check.

PR #1480, opened 2024-07-18
enha: check shutdown signal in tasks

marcospb19-cw:
Some long-living tasks weren't checking the shutdown signal.

This PR ensures those tasks check shutdown periodically.

This PR also removes receiver timeout warning logs where it wasn't important.


src/eth/miner/miner.rs
```diff
@@ -472,11 +473,24 @@ mod interval_miner {
...
-                return;
+                break;
             }
 
+            let tick = match ticks_rx.recv_timeout(Duration::from_secs(2)) {
+                Ok(tick) => tick,
+                Err(RecvTimeoutError::Disconnected) => break,
+                Err(RecvTimeoutError::Timeout) => {
+                    tracing::warn!(
+                        channel = stringify!(ticks_rx),
```
marcospb19-cw, line 486: Mmmm this log might not be necessary anymore, since we now have a custom log message (2 lines below).

PR #1498, opened 2024-07-20
draft: a helper binary `"deserialize-type"`

marcospb19-cw:


marcospb19-cw:
In the following months we'll be revisiting the discussion about changing the serialization format from `bincode` to something else.

After that is decided, we reconsider if we'd like to have a helper binary for deserialization.

PR #1530, opened 2024-07-24
feat: rocksdb: isolation and versioning

carneiro-cw:


marcospb19-cw:
Making it a draft again so nobody merges accidentally before the proper time.

PR #1594, opened 2024-08-02
chore: simplify log filter `to_block`

marcospb19-cw:
it was an `Option`, but unconditionally set as `Some`


marcospb19-cw:
I don't comprehend why this breaks subscriptions and therefore I'll choose to ignore and not to solve that TODO, if it breaks, maybe that TODO shouldn't be there.

PR #1658, opened 2024-08-21
chore: improve pr agent review on memory leak

gabriel-aranha-cw:


.github/workflows/pr-agent.yml
```diff
@@ -34,6 +34,7 @@ env:
     * When you detect an unwrap, suggest to use expect.
     * When you detect a core component or service is being initialized, ensure its initialization is being logged with all relevant configurations.
     * When too much cloning is happening, suggest to use references if possible.
+    * When you detect an infinite loop, ensure that tracing events inside the loop use `parent: None` to avoid unbounded Vec growth resulting in memory leak. Example: `tracing::info!(parent: None, "some message");`
```
mayconamaroCW, line 37: I believe these quotes won't break the env value because of YAML's `|`, but I'm commenting this to remember checking if it breaks.

mayconamaroCW (approved):

PR #1673, opened 2024-08-28
chore: bump sqlx version due to security alert

marcospb19-cw:
also had to bump tempfile for it to work


marcospb19-cw:
This version is broken, waiting for SQLX to release a new patch version with this fixed.

PR #1724, opened 2024-09-24
enha: Remove some unnecessary heap allocations in BlockchainClient

maxcabrajac:
I was having a look at the code base and found these quick and tiny one-line optimizations.
If binary sizes are a problem, then slices (`&[JsonValue]`) are a better choice than this.

### DIFF
--- a/src/infra/blockchain_client/blockchain_client.rs
+++ b/src/infra/blockchain_client/blockchain_client.rs
@@ -125,7 +125,7 @@ impl BlockchainClient {
     pub async fn fetch_listening(&self) -> anyhow::Result<()> {
         tracing::debug!("fetching listening status");
 
-        let result = self.http.request::<bool, Vec<()>>("net_listening", vec![]).await;
+        let result = self.http.request::<bool, _>("net_listening", [(); 0]).await;
         match result {
             Ok(_) => Ok(()),
             Err(e) => log_and_err!(reason = e, "failed to fetch listening status"),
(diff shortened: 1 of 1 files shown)


carneiro-cw:
You can run `just lint` fix the Lint and Style checks

carneiro-cw:
> You can run `just lint` fix the Lint and Style checks
> 
> edit by @mayconamaroCW: lint with nightly-2024-07-19 toolchain to ensure compatibility with CI

Otherwise lgtm

maxcabrajac:
Done

dinhani-cw (approved):

src/infra/blockchain_client/blockchain_client.rs
```diff
@@ -125,7 +125,7 @@ impl BlockchainClient {
     pub async fn fetch_listening(&self) -> anyhow::Result<()> {
         tracing::debug!("fetching listening status");
 
-        let result = self.http.request::<bool, Vec<()>>("net_listening", vec![]).await;
+        let result = self.http.request::<bool, _>("net_listening", [(); 0]).await;
```
marcospb19, line 128: FWIW empty Vec's aren't allocated in the heap.

maxcabrajac: Yeah I thought of that when writting, but didn't make it explicit. My intention was to  use the same pattern on every call for consistency and maybe to hint something like copilot not to use vecs in the future.

maxcabrajac: In fact, `Vec<()>` never allocates as () is zero-sized

PR #1752, opened 2024-10-10
feat: AccountTransfersEvent

dinhani-cw:
### **User description**
Schema proposal for transfers events to be emitted by the ledger.
Check each comment for specific concerns.

___

### **PR Type**
Enhancement

___

### **Description**
- Introduced `AccountTransfersEvent` struct to represent token transfers associated with a specific Ethereum account within a single transaction
- Defined `AccountTransfer` struct to represent individual token transfers between debit and credit parties
- Added new `ledger` module with `events` submodule to organize transfer-related structures
- Included comprehensive fields for event metadata, transaction details, and transfer specifics in the new structures
- Added UUID dependency to support unique identifiers for event publications
- Exposed the new `ledger` module in `lib.rs` for use throughout the project

___

### **Changes walkthrough** 📝
<tr>
  <td>
    <details>
      <summary><strong>events.rs</strong><dd><code>Define AccountTransfersEvent and AccountTransfer structs</code>&nbsp; </dd></summary>
<hr>

src/ledger/events.rs

<li>Introduced <code>AccountTransfersEvent</code> struct to represent token transfers <br>for a specific Ethereum account<br> <li> Defined <code>AccountTransfer</code> struct to represent individual token transfers<br> <li> Included fields for event metadata, transaction details, and transfer <br>specifics<br>

</details>

  </td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/1752/files#diff-3c32fd4a5d4c71b18e0d1b10b6f00f18c4dca2d3acd63d314343f7be6a219515">+95/-0</a>&nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>mod.rs</strong><dd><code>Add events module to ledger</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/ledger/mod.rs

- Added a public module declaration for `events`

</details>

  </td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/1752/files#diff-da1104bd92d4b1c43b6db870e58120fac09a3d3d80cb002d66d68ded27597a96">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
      <summary><strong>lib.rs</strong><dd><code>Expose ledger module in lib.rs</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/lib.rs

- Added `pub mod ledger;` to expose the new ledger module

</details>

  </td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/1752/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    
</table></td></tr><tr><td><strong>Dependencies</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>Cargo.toml</strong><dd><code>Add UUID dependency to Cargo.toml</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

Cargo.toml

- Added `uuid = "=1.10.0"` as a new dependency

</details>

  </td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/1752/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

___

> 💡 **PR-Agent usage**: Comment `/help "your question"` on any pull request to receive relevant information


src/ledger/events.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/// Represents a token transfer between a debit party and a credit party.
+pub struct AccountTransfer {
+    /// Address of the token contract that executed the transfer between `debit_party_address` and `credit_party_address`.
+    ///
+    /// It may differ from the main `contract_address` because any contract can execute transfers in token contracts.
+    ///
+    /// Format: Prefixed account address - 20 bytes - 0x1234567890123456789012345678901234567890
+    pub token_address: Address,
```
dinhani-cw, line 77: Token information (BRLC or JIM) can be derived from this field.

I need to validate if there is a change that non-token contracts can emit some form of transfer events, but I think that is not possible.

src/ledger/events.rs
```diff
@@ -0,0 +1,95 @@
...
+    ///
+    /// Format: Prefixed account address - 20 bytes - 0x1234567890123456789012345678901234567890
+    pub credit_party_address: Address,
+
+    /// Amount transferred from debit party to credit party.
+    ///
+    /// Format: TBD
+    /// * Prefixed hexadecimal amount - 32 bytes - 0x123450x1234567890123456789012345678901234567890678901234567890
+    /// * Decimal amount without prefix or padding - Integer - 1234
+    pub amount: U256,
```
dinhani-cw, line 94: Deciding the format of the amount is critical. Having it as decimal can make math operations in queries a lot easier without having to perform a lot of conversions from hexadecimal.

src/ledger/events.rs
```diff
@@ -0,0 +1,95 @@
...
+    ///
+    /// If the event is republished, a new ID will be generated for the publication.
+    ///
+    /// Format: UUID v7 (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx)
+    pub publication_id: Uuid,
+
+    /// Datetime of the event publication.
+    ///
+    /// Format: ISO 8601
+    pub publication_datetime: DateTime<Utc>,
```
dinhani-cw, line 27: Should we go with ISO 8601 or UNIX timestamps for datetimes?
I think ISO 8601 is more readable.

src/ledger/events.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Format: Integer
+    pub block_number: BlockNumber,
+
+    /// Datetime of the block that originated account transfers.
+    ///
+    /// Format: ISO 8601
+    pub block_datetime: DateTime<Utc>,
+
+    /// List of transfers the related `account_address` is part of.
+    pub transfers: Vec<AccountTransfer>,
```
dinhani-cw, line 67: If necessary, this can be split in two arrays (credits and debits).

`credits` would contains only transactions where the account_address is the credit party and `debits` where it is the debit party.

But the current approach with a single array may work better if expanding the array in multiple rows (unnest operation).

src/ledger/events.rs
gabriel-aranha-cw: I think ISO is better as well. It is easier to understand and no need to convert anything. We want to facilitate the consumers of these events, which will be ourselves too in case of debugging.

src/ledger/events.rs
gabriel-aranha-cw: Agreed on decimal format.

src/ledger/events.rs
gabrielpioto: I prefer ISO 8601 as well, but if message size is a constraint I believe unix timestamp is better

src/ledger/events.rs
gabrielpioto: I think single array is the way to go.

src/ledger/events.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/// Represents a token transfer between a debit party and a credit party.
+pub struct AccountTransfer {
```
gabrielpioto, line 71: I believe this format of debit_party and credit_party works well, but my main concern is that we will inevitably need to infer a some of information from them.
For instance: We will need to compare, for each transfer, both addresses with `AccountTransferEvent.account_address` to determine the direction of the transfer (credit or debit)
Also, the one that is not equal to the account_address will determine the type of transfer.

I believe it would be nice if these two inferences were already done by the producer

src/ledger/events.rs
gabrielpioto: +1 for decimal format

src/ledger/events.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::eth::primitives::BlockNumber;
+use crate::eth::primitives::Hash;
+
+/// Represents token transfers (debits and credits) associated with a specific Ethereum account within a single transaction.
+///
+/// The `account_address` field identifies the main account involved in these transfers.
+///
+/// An event will be generated for each account involved in a transaction, meaning if a transaction is not the main account in this event,
+/// in another event it will be and credit and debit parties will be adjusted accordingly.
+pub struct AccountTransfersEvent {
```
gabrielpioto, line 16: It always nice to have a schema_version field. Even if we don't need now and don't think will gonna need, better safe than sorry :) 

PS. maybe we can use kafka headers for this, don't know

src/ledger/events.rs
gabrielpioto, line 71: Shouldn't there be a datetime field here as well?

dinhani-cw: Good point.
I think I will add a `kind` field inside each transfer with the possible values being `debit` and `credit`?
`debit` means the `account_address` is the `debit_party`, `credit` means it is the `credit_party`.

dinhani-cw: It would be the same as the `block_datetime`. Should we rename block_datetime to make it more clear?

vicenzi-cw: @gabrielpioto do you have an example of how it would be built?

src/ledger/events.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub publication_datetime: DateTime<Utc>,
+
+    /// Idempotency key of the event payload (TBD: `account_address` + `transaction_hash`).
+    ///
+    /// It is unique for each distinct event, but consistent across retries or republications of the same event payload.
+    ///
+    /// Format: String (TBD)
+    pub idempotency_key: String,
+
+    /// Address of the account that is part of all transfers.
```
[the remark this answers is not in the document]
dinhani-cw: It means that this account is present in all the transfers in the transfers array. It is the event subject. All values are about it.

src/ledger/events.rs
vicenzi-cw: I guess so, @dinhani-cw . Doesn't matter much from where you're deriving the `transaction_datetime`, I think

src/ledger/events.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// Address of the contract that originated transfers.
+    ///
+    /// Format: Prefixed account address - 20 bytes - 0x1234567890123456789012345678901234567890
+    pub contract_address: Address,
+
+    /// Identifier of the Ethereum function that originated transfers.
+    ///
+    /// Format: Prefixed function signature - 4 bytes - 0x12345678
+    pub function_id: [u8; 4],
```
vicenzi-cw, line 54: what do you think to receive this as formatted hex as well?

src/ledger/events.rs
gabrielpioto: @dinhani-cw I prefer the term `direction` over `kind`. In my opinion `kind` is "too generic"
@vicenzi-cw What do you mean? For `direction` we have all information we need to infer in the message. For the type we will need a mapping somewhere of addresses and the event_type. Maybe as a env var / config file

dinhani-cw: Added it as `direction`.

src/ledger/events.rs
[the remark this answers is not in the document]
dinhani-cw: @vicenzi-cw Contracts are accounts. All transfers have two accounts as defined [here](https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#IERC20-Transfer-address-address-uint256-).

Internal movements inside a contract are not transfer and are not covered by these events.

src/ledger/events.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    /// Datetime of the Ethereum transaction that originated transfers.
+    ///
+    /// Format: ISO 8601
+    pub transaction_datetime: DateTime<Utc>,
+
+    /// Address of the contract that originated transfers.
+    ///
+    /// Format: Prefixed account address - 20 bytes - 0x1234567890123456789012345678901234567890
+    pub contract_address: Address,
```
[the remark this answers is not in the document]
dinhani-cw: Only if it is a transfer of native currency (Wei), but that is not being covered here and it is not used in the company.
If the transfer is from BRLC or JIM, it always involve a contract.

src/ledger/events.rs
vicenzi-cw: sry, you can ignore my comment, @gabrielpioto . I was thinking about single-entry and made some comments that doesnt make sense

src/ledger/events.rs
carneiro-cw: Just as a clarification, `account_address` here is the `primary_account_address`? 
If so I think we should decide on one name to use throughout the documentation/code as to avoid any confusion.

src/ledger/events.rs
carneiro-cw: I think having the fields be named `from` and `to` instead of `credit_party_address` and `debit_party_address` is just as clear and more consistent with the language used in other places, such as in the Transfer event emitted by ERC20 for instance.
I'm not arguing we should be stuck to blockchain lingo, but in this case I find it more readable (I might be biased since we've spent this year focused in developing stratus)

src/ledger/events.rs
carneiro-cw: I think it is "possible", someone, someday might create a contract that emits its own Transfer event with the same signature, I don't think there is anything that would bar someone from doing so if they wanted.

Either way, we could filter those logs out since we know the addresses of JIM and BRLC. That way even if someone, someday for some reason decided that it would be a good idea to emit an event in the same format of the ERC20 Transfer event it wouldn't matter because we would know to only publish events that concerned either contract.

src/ledger/events.rs
dinhani-cw: @carneiro-cw Yes, `account_address` was the original name before I renamed it to `primary_account_address`. This was just a comment I missed renaming. Now it is fixed.

src/ledger/events.rs
dinhani-cw: @carneiro-cw As decided together with the statements team, the implementation in Stratus will not use these mappings we have **for events**, that includes determining if the contract is JIM or BRLC.

This way they don't have to wait for a Stratus change and deployment in case these addresses changes. So they will have to handle on their side if something is BRLC, JIM or something else.

So if anything emits an event with the Transfer signature, it will be handled as a transfer.

src/ledger/events.rs
dinhani-cw: @carneiro-cw 

Credit party and debit party are common names used by some financial institutions.
* [Pismo](https://developers.pismo.io/events/docs/integrated-payments-transfer-out-created-1)
* [BTG](https://developer.btgpactual.com/doctr#section/External-Transfer-Status)

Consider the scenario where we want to evolve the API and publish additional information like:
* `debit_party_account_type` / `credit_party_account_type` (normal account or contract account)
* `debit_party_previous_balance` /  `credit_party_previous_balance` (balance before operation)

It would feel more natural than having it as `from_account_type`, `from_previous_balance`, `to_account_type` or `to_previous_balance`.

carneiro-cw: I imagined they were common names, but I think from cloudwalk's POV, since we already implemented the whole transactional structure on top of the blockchain, From and To are even more common.

As for the second point: If the data is flattened I agree. But If it is structured (such as in the BTG example you linked), I think having something like:
```
"from": {
      "account": "000000",
      "accountType": "CC",
      "bank": "208",
      "branch": "1",
      "name": "BTG Pactual",
      "personType": "J",
      "taxId": "00000000000000"
    },
```
Still feels natural (to me). So I stand on my point of preserving naming consistency as much as we can.

Either way I think the naming is appropriate, I just fear that having multiple terms to mean the same thing will cause points of confusion, specially with those less involved in the development of the system's that decide the terms.

marcospb19-cw: To me `from_account_type` is more natural than `debit_party_account_type`, even if it's grammatically weird.

But over both, I prefer `sender`, `receiver`, `sender_account_type` and `receiver_account_type`.

carneiro-cw: In this case I concede that credit_party_<suffix> and debit_party_<suffix> reads more naturally, and since it is an industry standard it does makes sense to use it.
(I still like `from` and `to` better though :p mainly for naming consistency across different systems)

src/ledger/events.rs
```diff
@@ -0,0 +1,109 @@
...
+    ///
+    /// Format: ISO 8601
+    pub publication_datetime: DateTime<Utc>,
+
+    /// Idempotency key of the event payload (TBD: `primary_account_address` + `transaction_hash`).
+    ///
+    /// It is unique for each distinct event, but consistent across retries or republications of the same event payload.
+    ///
+    /// Format: String (TBD)
+    pub idempotency_key: String,
```
marcospb19-cw, line 36: I'd call this "event_id" and put it at the top, above "publication_id", I assume it's what clients really care about, over publication id.

dinhani-cw:
Update: this proposal aims to use a flattened struct, so clients does not have to deal with too much nesting.

src/ledger/events.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use crate::eth::primitives::BlockNumber;
+use crate::eth::primitives::Hash;
+
+/// Represents token transfers (debits and credits) associated with a specific Ethereum account within a single transaction.
+///
+/// The `primary_account_address` field identifies the primary account involved in these transfers.
+///
+/// An event will be generated for each account involved in a transaction, meaning if a transaction is not the primary in this event,
+/// in another event it will treated as the primary and credit and debit operations adjusted accordingly.
```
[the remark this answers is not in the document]
dinhani-cw: Stratus will use only the `Transfer` events emitted by ERC-20 contracts.

It will generate one event payload per (transaction hash + account) and not one payload event per transaction, so batch transactions does not really changes anything.

src/ledger/events.rs
dinhani-cw: I did consider calling it just `id` or `event_id`, but some internal systems like Issuing already uses `idempotency_key` for similar scenarios. Some public APIs like Stripe and other payment providers also have fields with some variation of idempotency key.

dinhani-cw:
Update 2: Fields can be renamed before the final release. Please focus on more critical decisions like missing fields, value formats and other ledger domain related rules.

src/ledger/events.rs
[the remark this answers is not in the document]
marcospb19-cw: So for a batch transaction with size 5, we'd emit 6 events:

- 1 with where the primaryAccount is the debitParty.
- 5 where the primaryAccount is the creditParty.

Got it, thanks for clarifying!

gabriel-aranha-cw:
Final benchmark:
RPS Stats: Max: 868.00, Min: 651.00, Avg: 816.68, StdDev: 26.23
TPS Stats: Max: 868.00, Min: 603.00, Avg: 816.34, StdDev: 28.46
Plot: [View Plot](https://storage.cloud.google.com/stratus_performance_benchmark_staging/main/pr1752_main_bd0c312_bench-555220731.png)

PR #1775, opened 2024-10-18
feat: event amount serialized with rust_decimal and divided by 1.000.000

dinhani-cw:


gabriel-aranha-cw:
Final benchmark:
RPS Stats: Max: 919.00, Min: 788.00, Avg: 838.12, StdDev: 21.61
TPS Stats: Max: 886.00, Min: 788.00, Avg: 837.79, StdDev: 20.74
Plot: [View Plot](https://storage.cloud.google.com/stratus_performance_benchmark_staging/main/pr1775_main_0ce3c97_bench-3495742319.png)

PR #1888, opened 2024-11-28
enha: remove mutex that keeps transactions from executing while a block is being commited

carneiro-cw:
### **PR Type**
Enhancement

___

### **Description**
- Removed mutex that prevents transactions from executing while a block is being committed
- Implemented EVM input validation to ensure transaction input matches block header
- Added new error type `TransactionEvmInputMismatch` for handling mismatched EVM inputs
- Updated `LocalTransactionExecution` to include `evm_input` field
- Added new test contract `TestEvmInput` and corresponding E2E tests for EVM input validation
- Modified CI workflow to include interval-based Stratus testing
- Updated `justfile` to support interval-based block mode in E2E tests
- Added derive macros to `EvmInput` and `PointInTime` structs for improved serialization and testing support

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>6 files</summary><table>
<tr>
  <td>
    <details>
      <summary><strong>evm_input.rs</strong><dd><code>Add derive macros to EvmInput struct</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/eth/executor/evm_input.rs

<li>Added <code>serde::Deserialize</code>, <code>fake::Dummy</code>, and <code>PartialEq</code> derive macros to <br><code>EvmInput</code> struct<br>

</details>

  </td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/1888/files#diff-9dbdf7472d01464e439067cbd23dfe3648c7fb38a307bee0cf8642ad16a1a252">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>executor.rs</strong><dd><code>Refactor transaction execution and error handling</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/eth/executor/executor.rs

[...]

PR #1887, opened 2024-11-28
enha: implement storage cache

marcospb19-cw:
### **PR Type**
Enhancement

___

### **Description**
- Implemented a new StorageCache system using LRU caches for slots and accounts to improve performance.
- Refactored StratusStorage to integrate the new caching mechanism, modifying read_account and read_slot methods.
- Updated the importer_offline logic with new constants for cache size and batch management.
- Enhanced the Address struct with additional ordering traits.
- Replaced InMemoryTemporaryAccount with the more versatile AccountWithSlots struct.
- Added new dependencies (lru, parking_lot, rustc-hash, smallvec) to support the caching implementation.
- Updated e2e tests to reflect changes in error messages.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>importer_offline.rs</strong><dd><code>Refactor block execution and storage saving logic</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/bin/importer_offline.rs

<li>Replaced STORAGE_SAVER_BATCH_SIZE with new constants for cache size <br>and batch management<br> <li> Updated channel capacity for execute_to_save_tx<br> <li> Modified block execution loop to use new SAVER_BATCH_SIZE<br>

</details>

  </td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/1887/files#diff-9fef7d584818c3db119ea1cd0952a57a953751f85f93ca813af861c4c737c53a">+12/-16</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>address.rs</strong><dd><code>Add ordering traits to Address struct</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/eth/primitives/address.rs

- Added PartialOrd and Ord traits to Address struct derivation

</details>

  </td>
[...]

PR #1904, opened 2024-12-06
enha + fix: possible storage bug and remove states vec from inmemory temporary storage

carneiro-cw:
With the introduction of the storage cache #1887 we made it so the inmemory temporary storage only held one block in its block Vec. However this combined with #1888 introduces a possibly critical bug. When the pending block is evicted from the temporary storage it takes a few ms to be committed to the permanent storage, however executions can still happen in the interim and if an execution tries to read a slot/account that has not yet been committed to the permanent but has already been evicted from the temporary it would get the wrong value.

The cache implementation masks this bug pretty well, since all the recent changes will probably be in the cache, however it is not guaranteed that it won't evict some slot/account before we have committed the block to the permanent storage, and therefore we should not rely on it for this purpose.

This PR makes it so the temporary storage holds 2 blocks, the pending block, and the latest to avoid this issue.

### DIFF
--- a/src/eth/storage/temporary/inmemory.rs
+++ b/src/eth/storage/temporary/inmemory.rs
@@ -2,10 +2,6 @@
 
 use std::collections::HashMap;
 use std::sync::RwLock;
-use std::sync::RwLockReadGuard;
-use std::sync::RwLockWriteGuard;
-
-use nonempty::NonEmpty;
 
 use crate::eth::executor::EvmInput;
 use crate::eth::primitives::Account;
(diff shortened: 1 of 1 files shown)


marcospb19-cw (dismissed):

gabriel-aranha-cw:
Final benchmark:
Run ID: bench-1557210719

Git Info:
- Branch: main
- Commit: d1c86fe
- PR: #1904

Configuration:
- Target Account Strategy: Default

RPS Stats: Max: 1244.00, Min: 767.00, Avg: 1099.94, StdDev: 70.23
TPS Stats: Max: 1236.00, Min: 866.00, Avg: 1066.76, StdDev: 75.87

Plot: [View Plot](https://storage.cloud.google.com/stratus_performance_benchmark_staging/main/pr1904_main_d1c86fe_bench-1557210719.png)

PR #2070, opened 2025-03-26
Rocks revert to blocks with tests

arthurmm-cloudwalk:
Redefines the rocks revert code ...
* With binary and option through rpc_server in dev mode for tests
* Tests of revert blocks


carneiro-cw:
Stale and lots of conflicts, closing

PR #2308, opened 2025-09-09
chore: remove execution changes from slot map

carneiro-cw:


carneiro-cw:
closed in favor of #2309

PR #2520, opened 2026-05-28
feat: same signer recovery path

f3l1ph3s:
### **PR Type**
Enhancement

___

### **Description**
- Add `recover_signer` method for consistent ECDSA signer recovery

- Update RLP decoding to invoke `recover_signer` on deserialization

- Modify `TryFrom` conversions to call `recover_signer`

- Remove direct signer extraction in `try_from_alloy_transaction`

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["TxEnvelope"] --> B["try_from_alloy_transaction (placeholder signer)"]
  B --> C["TransactionInput"]
  C -- "recover_signer" --> D["TransactionInput with actual signer"]
```

<tr>
  <td>
    <details>
      <summary><strong>transaction_input.rs</strong><dd><code>Add unified ECDSA signer recovery</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/eth/primitives/transaction_input.rs

<ul><li>Added <code>recover_signer</code> method for ECDSA recovery<br> <li> Updated <code>Decodable::decode</code> to call <code>recover_signer</code><br> <li> Modified <code>TryFrom<ExternalTransaction></code> and <code>TryFrom<AlloyTransaction></code> to invoke recovery<br> <li> Removed inline signer extraction in <code>try_from_alloy_transaction</code>, using <br>placeholder ZERO</ul>

</details>

  </td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2520/files#diff-82805cb38a9fb8fc686e7186838bde829d2e092c13ec88636aa57a4129ffedde">+35/-15</a>&nbsp; </td>

</tr>

</details>

___


src/eth/primitives/transaction_input.rs
```diff
@@ -164,7 +183,7 @@ fn try_from_alloy_transaction(value: alloy_rpc_types_eth::Transaction) -> anyhow
         execution_info: ExecutionInfo {
             chain_id: value.inner.chain_id().map(Into::into),
             nonce: Nonce::from(value.inner.nonce()),
-            signer,
+            signer: Address::ZERO,
```
carneiro-cw: In this case I think it'd be better to have a Signer { Recovered(Address), Unrecovered }  and add a signer() accessor to TransactionInput that either returns the signer address or recovers then returns.

carneiro-cw: It would need to get a mutable reference to do that though, idk if this would work everywhere. Either way it'd be better to find an alternative to placeholder values .

carneiro-cw (changes requested):
We can add tests to avoid regressions for this. In the e2e testing suite try creating each existing transaction type and using all fields that alloy can use to create a signature. Right now we're blocking certain tx types though, so maybe put the conditional that blocks the tx types in rpc_server behind a #cfg[not(dev)]. This tx would need to be sent to the leader and imported to the follower, so leader-follower tests are the right place to do this.

src/eth/primitives/transaction_input.rs
```diff
@@ -80,19 +80,15 @@ pub struct TransactionInput {
...
     ///
     /// This must be called after constructing a TransactionInput to ensure the signer
-    /// is derived from the same canonical fields regardless of the origin.
-    pub fn recover_signer(&mut self) -> anyhow::Result<()> {
-        let alloy_tx: AlloyTransaction = self.clone().into();
-        let signer_raw = alloy_tx
-            .inner
-            .inner()
+    /// is derived from the same recovery path regardless of the origin.
+    fn recover_signer(&mut self, envelope: &TxEnvelope) -> anyhow::Result<()> {
```
carneiro-cw: I'm pretty sure this is a regression from the previous commit in this PR, and equivalent to what we already do now. 
The whole point is that there are fields that we do not store in stratus, but the leader will use to recover the signer since its recovering from the original transaction that we parsed directly to Alloy.
The follower will not have the original TxEnvelope so it will create a different signer.
The AI review was correct that this might cause the leader to generate a wrong signer according to ethereum spec if it includes fields we don't support, but that is preferable over not being able to recover the signer after we commit the transaction to storage.
We'll need to reject any fields we don't support instead of just parsing the RLP to an Alloy transaction then converting it to TransactionInput to address this other issue, but this is outside of the scope of this PR.

carneiro-cw (changes requested):

e2e/cloudwalk-contracts/integration/test/helpers/account.ts
```diff
@@ -83,6 +83,100 @@ export class Account implements Addressable {
...
+            accessList: [
+                {
+                    address: "0x0000000000000000000000000000000000000001",
+                    storageKeys: ["0x0000000000000000000000000000000000000000000000000000000000000001"],
+                },
+            ],
+            maxFeePerBlobGas: 1000000000,
+            blobVersionedHashes: [],
+        });
+    }
```
carneiro-cw, line 171: you can test type 4 transactions too

carneiro-cw: To do that you can add #[cfg(feature = "dev")] to this [conditional](https://github.com/cloudwalk/stratus/blob/28beb3d827858c8b3840b82965acdccef58713cd/src/eth/rpc/rpc_server.rs#L1273)

e2e/cloudwalk-contracts/integration/test/leader-follower-tx-types.test.ts
```diff
@@ -0,0 +1,171 @@
...
+        const followerEIP2930Receipt = await sendWithRetry("eth_getTransactionReceipt", [eip2930Hash]);
+        expect(followerEIP2930Receipt.from.toLowerCase()).to.equal(DAVE.address.toLowerCase());
+
+        console.log("Getting receipt for Type 2 (EIP-1559) tx:", eip1559Hash);
+        const followerEIP1559Receipt = await sendWithRetry("eth_getTransactionReceipt", [eip1559Hash]);
+        expect(followerEIP1559Receipt.from.toLowerCase()).to.equal(CHARLIE.address.toLowerCase());
+
+        console.log("Getting receipt for Type 3 (EIP-4844) tx:", eip4844Hash);
+        const followerEIP4844Receipt = await sendWithRetry("eth_getTransactionReceipt", [eip4844Hash]);
+        expect(followerEIP4844Receipt.from.toLowerCase()).to.equal(EVE.address.toLowerCase());
```
carneiro-cw: NIT: For thix PR the objective was getting the same from, but the entire receipt should be equal, I think there is some utility to deep compare full objects and using that might be more appropriate.

carneiro-cw: If there isn't maybe we should add one (?)

src/eth/executor/evm_input.rs
```diff
@@ -87,7 +87,7 @@ impl EvmInput {
     /// Creates from a transaction that was sent to Stratus with `eth_sendRawTransaction` or during Importing.
     pub fn from_eth_transaction(input: &ExecutionInfo, block_number: BlockNumber, block_timestamp: UnixTime) -> Self {
         Self {
-            from: input.signer,
+            from: input.signer.address().unwrap(),
```
carneiro-cw: No unwrap, have the function recover the signer and return that in case the signer is unrecovered, that means that the accessor used here will have to be in the input where the signer can be computed.

src/eth/primitives/transaction_input.rs
```diff
@@ -79,21 +126,49 @@ pub struct TransactionInput {
...
 }
 
+impl TransactionInput {
+    /// Returns the recovered signer address.
+    ///
+    /// # Panics
+    /// Panics if the signer has not been recovered. All construction paths
+    /// call `recover_signer` before returning, so this is safe in practice.
+    pub fn signer(&self) -> Address {
+        self.execution_info.signer.address().unwrap()
```
carneiro-cw: As I mentioned in another comment, do not unwrap, recover the signer and return it in case address() is unrecovered, maybe add a warning log so we're aware that there is some path that is not calling recover_signer before returning.

src/eth/primitives/transaction_input.rs
```diff
@@ -79,21 +126,49 @@ pub struct TransactionInput {
...
                 block_hash: None,
                 block_number: None,
                 transaction_index: None,
                 effective_gas_price: None,
             })
-            .map_err(|_| rlp::DecoderError::Custom("failed to convert transaction"))
+            .map_err(|_| rlp::DecoderError::Custom("failed to convert transaction"))?;
+
+            tx_input
+                .recover_signer(&envelope)
```
carneiro-cw: you're still recovering the signer from the original envelope. Remove the envelope field from recover_signer and recover the signer from the fields we actually have in TransactionInput. Since we do rely on alloy to recover signers currently, this will involve a sort of redundant conversion from transactioninput to alloytransaction after we just converted from transactioninput.

src/eth/primitives/transaction_input.rs
```diff
@@ -79,21 +126,49 @@ pub struct TransactionInput {
...
+    /// call `recover_signer` before returning, so this is safe in practice.
+    pub fn signer(&self) -> Address {
+        self.execution_info.signer.address().unwrap()
+    }
+
+    /// Recovers the signer from the original transaction envelope using ECDSA recovery.
+    ///
+    /// This must be called after constructing a TransactionInput to ensure the signer
+    /// is derived from the same recovery path regardless of the origin.
+    fn recover_signer(&mut self, envelope: &TxEnvelope) -> anyhow::Result<()> {
```
carneiro-cw: This should only have `&mut self` as a parameter.

carneiro-cw (changes requested):

carneiro-cw (changes requested):
As I said before, we won't be updating the rocks types for this. Instead we'll compute the singer ignoring the fields we don't support for now.

e2e/cloudwalk-contracts/integration/test/leader-follower-tx-types.test.ts
```diff
@@ -0,0 +1,209 @@
...
+
+import { ALICE, BOB, CHARLIE, DAVE, EVE, FERDIE } from "./helpers/account";
+import {
+    sendAndGetFullResponse,
+    sendWithRetry,
+    updateProviderUrl,
+    waitForFollowerToSyncWithLeader,
+} from "./helpers/rpc";
+
+describe("Leader & Follower transaction types signer recovery regression test (fields supported by Stratus)", function () {
```
carneiro-cw: not really a regression test if you're not sending the unsuported fields, I'd expect the "fullFields" functions to create the txns with all possible fields that tx can have.

e2e/cloudwalk-contracts/integration/test/helpers/account.ts
```diff
@@ -83,6 +83,106 @@ export class Account implements Addressable {
...
+            to: counterParty,
+            value: amount,
+            chainId: CHAIN_ID_DEC,
+            maxFeePerGas: 2000000000,
+            maxPriorityFeePerGas: 2000000000,
+            gasLimit,
+            nonce,
+            type: 4,
+            data: "0xdeadbeef",
+            accessList: [],
```
carneiro-cw: add something to the extra fields

e2e/cloudwalk-contracts/integration/test/helpers/account.ts
```diff
@@ -83,6 +83,106 @@ export class Account implements Addressable {
...
+        return await this.signer().signTransaction({
+            to: counterParty,
+            value: amount,
+            chainId: CHAIN_ID_DEC,
+            gasPrice: 2000000000,
+            gasLimit,
+            nonce,
+            type: 1,
+            data: "0xdeadbeef",
+            accessList: [],
```
carneiro-cw: don't send empty fields for these tests

e2e/test/helpers/rpc.ts
```diff
@@ -626,7 +626,7 @@ export async function createEIP1559Transaction(account: Account) {
         to: BOB.address,
         value: toHex(TEST_TRANSFER),
         maxFeePerGas: toHex(2),
-        maxPriorityFeePerGas: toHex(1),
+        maxPriorityFeePerGas: toHex(2),
```
carneiro-cw: why?

carneiro-cw (changes requested):

carneiro-cw (changes requested):
You have to add the tests to the matrix in the github workflow, It is not running in the CI

carneiro-cw (dismissed):

carneiro-cw (approved):

PR #2547, opened 2026-07-19
feat(rpc): implement debug_traceCall endpoint

### LINKED ISSUE #2539: debug_traceCall endpoint
Similar to `debug_traceTransaction` but without having to send the transaction before calling it.

Thiago316316:
## What

Implements `debug_traceCall` (issue #2539) — a traced version of `eth_call`.
It runs an arbitrary call against a chosen point-in-time state and returns a
Geth-style trace, **without** the call having to be signed or mined first
(i.e. a dry-run simulation of `debug_traceTransaction`).

Signature: `debug_traceCall(call, block, options)` where
- `call` — `CallInput` (`from`/`to`/`value`/`data`),
- `block` — `BlockFilter` (`latest`/`pending`/`earliest`/hash/number),
- `options` — `GethDebugTracingOptions` (tracer selection).

## How

- **`evm_input.rs`** — `InspectorInput` becomes an enum with `Transaction` and
  `Call` variants (previously a transaction-only struct).
- **`evm.rs`** — `inspect` dispatches on the variant. Transaction logic is
  unchanged (renamed to `inspect_transaction`); the tracer match block is
  extracted into a shared `run_tracer`. New `inspect_call` builds an `EvmInput`
  from the pending/mined block with no tx-hash lookup and no block replay.
- **`executor.rs`** — new `trace_call` mirrors `trace_transaction` and reuses
  the existing `inspector` worker pool (`InspectorTask`), so span propagation
  across the worker-thread boundary is inherited unchanged.
- **`rpc_server.rs`** — registers `debug_traceCall`, parses the three params,
  resolves the block to a point-in-time, traces, and decorates the result.

## Testing

- `just test` — full Rust suite green (incl. the pre-existing
  `debug_traceTransaction` coverage, which shares `run_tracer`).
- `just lint-check` — clean.
- New e2e cases in `e2e-json-rpc.test.ts`:
  - `callTracer` traces a contract call that was never sent (latest),
  - traces against pending state without mutating it.
- Manual verification of every tracer type against a running node.

## Compatibility

`debug_traceTransaction` is unchanged behaviorally — it takes the same
`run_tracer` path; only the input plumbing was refactored.

Closes #2539


Thiago316316:
Hi @carneiro-cw  i already got a positive review on the agent, if you have some time to see would be good. If there is something that needs to change, just say it. Thanks!!

PR #2653, opened 2026-08-28
feat: decode tx_input directly from RLP

f3l1ph3s:
### **PR Type**
Enhancement, Tests

___

### **Description**
- Add `signature_hash()` method to `TransactionInput`
  - Supports Legacy and EIP-2930/1559/4844/7702 types

- Refactor RLP decoding flow
  - Introduce `build_transaction_input_from_envelope()`
  - Replace `try_from_alloy_transaction()` logic

- Remove redundant transaction conversion code

- Add unit tests for `signature_hash()` correctness

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["RLP bytes"] -- "Decodable.decode()" --> B["TxEnvelope"]
  B -- "build_transaction_input_from_envelope()" --> C["TransactionInput"]
  C -- "signature_hash()" --> D["B256 signature hash"]
```

<tr>
  <td>
    <details>
      <summary><strong>transaction_input.rs</strong><dd><code>Add signature_hash and refactor decoding</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/eth/types/transaction/transaction_input.rs

<ul><li>Introduce <code>signature_hash()</code> computing B256 from stored fields<br> <li> Refactor RLP decoding into <code>build_transaction_input_from_envelope()</code><br> <li> Replace old <code>try_from_alloy_transaction()</code> conversion path<br> <li> Add unit tests verifying signature hash matches envelope</ul>

</details>

  </td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2653/files#diff-d208d847e351c20308e645f86113221880176118dfae5c3f1b13f57a85c28aea">+144/-26</a></td>

</tr>

</details>

___


carneiro-cw (changes requested):
Wrong idea. The idea is to decode transaction input directly from the received rlp. This means that we never even create the alloy types in the first place, the rlp gets decoded directly to `TransactionInput`, this means we don't ever even decode to `TxEnvelope`.

Before doing this however, a better issue to solve first is #2548 . Since then when receiving an `eth_sendRawTransaction` we wouldn't have to do any conversions with alloy at all.

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +383,239 @@ impl TransactionInput {
...
+            27 => Ok((None, 0)),
+            28 => Ok((None, 1)),
+            v if v >= 35 => {
+                let chain_id = (v - 35) / 2;
+                let parity = (v - 27) % 2;
+                Ok((Some(ChainId::from(chain_id)), parity))
+            }
+            _ => Err(alloy_rlp::Error::Custom("invalid legacy v value")),
+        }
+    }
```
gventino-cw, line 433: a lot of magic numbers, maybe extract them to consts

src/eth/types/transaction/transaction_input.rs
```diff
@@ -143,10 +150,134 @@ impl TransactionInput {
...
+                    &AccessList::default(),
+                ]);
+                let mut out = Vec::with_capacity(1 + encoded.len());
+                out.push(0x01);
+                out.extend_from_slice(&encoded);
+                B256::from(keccak256(out))
+            }
+
+            // EIP-1559
+            2 => {
```
gventino-cw: maybe create an enum for this tx_types to improve readability (if one already doesn't exist yet), then u can remove this comments

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +383,239 @@ impl TransactionInput {
...
+        let to: Option<Address>;
+        let value: U256;
+        let input: Vec<u8>;
+        let v: u64;
+        let r: U256;
+        let s: U256;
+
+        match tx_type {
+            // EIP-2930
+            1 => {
```
gventino-cw: same as previous comment here

src/eth/types/transaction/transaction_input.rs
```diff
@@ -143,10 +150,134 @@ impl TransactionInput {
...
+                    &chain_id,
+                    &nonce,
+                    &gas_price,
+                    &gas_limit,
+                    &to.as_slice(),
+                    &value,
+                    &input.as_slice(),
+                    &AccessList::default(),
+                ]);
+                let mut out = Vec::with_capacity(1 + encoded.len());
```
gventino-cw: u're repeating this snippet at every match, the only thing that it's changing it's the hex value that u're passing to `out.push`

maybe it's worth extracting this snippet out of the match statement and only apply the `out.push` at the match

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +383,239 @@ impl TransactionInput {
...
+                let to_addr: AlloyAddress = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing to"))?;
+                to = Some(Address::from(to_addr.0));
+                value = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing value"))?;
+                let input_bytes: AlloyBytes = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing input"))?;
+                input = input_bytes.to_vec();
+                let _: AccessList = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing accessList"))?;
+                let _: Vec<SignedAuthorization> = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing authorizationList"))?;
+                v = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing v"))?;
+                r = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing r"))?;
+                s = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing s"))?;
```
gventino-cw: these three are very similar, maybe there's a way to u abstract/generalize this snippet of code

src/eth/types/transaction/transaction_input.rs
gventino-cw: noticed that u're also adding new fields to the array passed into `encode_rlp_list` depending of the EIP, my suggestion stands

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +383,239 @@ impl TransactionInput {
...
+
+        let nonce: u64 = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing nonce"))?;
+        let gas_price: u128 = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing gasPrice"))?;
+        let gas_limit: u64 = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing gasLimit"))?;
+        let to_bytes: AlloyBytes = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing to"))?;
+        let value: U256 = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing value"))?;
+        let input: AlloyBytes = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing input"))?;
+        let v: u64 = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing v"))?;
+        let r: U256 = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing r"))?;
+        let s: U256 = rlp.get_next()?.ok_or(alloy_rlp::Error::Custom("missing s"))?;
```
gventino-cw: maybe u can implement `from` or `try_from` to do this in a more self-contained way

gventino-cw: to do it u would probably need to create a dto with those fields

at stratus we usually don't do that if the dto will be barely used, so take ur time and think about it 

If u can define a good, small dto and it will be well reused at the module, it's ok

src/eth/types/transaction/transaction_input.rs
f3l1ph3s: i love magic numbers ✨

gventino-cw: images

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +383,239 @@ impl TransactionInput {
...
+        };
+
+        let signer = tx.recover_signer_address()?;
+        tx.execution_info.signer = Signer::Recovered(signer);
+
+        Ok(tx)
+    }
+
+    /// Decodes a legacy transaction from raw RLP bytes.
+    fn decode_legacy(raw_bytes: &[u8]) -> alloy_rlp::Result<Self> {
```
carneiro-cw: Use stratus error types (`src/eth/rpc/types/error.rs`), either return `StratusError` or some of the more specific enums. If you think it makes sense you can create a new error enum for RLP parsing and add that as a variant in `StratusError` OR in `RpcError` (Maybe changing `RpcError::ParameterDecodeError` but idk maybe some new variant)

src/eth/types/transaction/transaction_input.rs
carneiro-cw: The right approach would be to implement `Decodable` for each of the the types in `TransactionInput`. Most, if not all should be pretty straightforward as we'd only need to delegate to the newtype's inner Decodable implementation.

It might be possible to make a `trait Primitive` that has a type parameter `type inner: Decodable` and then `impl<T: StratusDecodable> Decodable for T` where the decode function returns `Self(inner::decode(buf))`. Or if that doesn't work, a derive macro in stratus_macros.

Or maybe just use the `RlpDecodable`/`RlpEnecodable` derive macros from alloy_rlp. I think it does that in a more generalized manner.

src/eth/types/transaction/transaction_input.rs
carneiro-cw: Also use anyhow::Result

src/eth/types/transaction/transaction_input.rs
carneiro-cw: This is dead code now :) you can remove it

src/eth/types/transaction/transaction_input.rs
```diff
@@ -310,9 +638,9 @@ impl TryFrom<AlloyTransaction> for TransactionInput {
     }
 }
 
-fn try_from_alloy_transaction(value: alloy_rpc_types_eth::Transaction) -> anyhow::Result<TransactionInput> {
+fn build_transaction_input_from_envelope(envelope: &TxEnvelope) -> anyhow::Result<TransactionInput> {
```
carneiro-cw, line 677: This is still needed for now because of the importer .

I think today one easy win would be updating `stratus_get_block_and_receipts` to serialize `stratus::eth::types::Block` directly instead of converting to the alloy type. Then updating the importers. We can leave this to another PR but should be straightforward. We probably should add an optional parameter to the request for selecting what type of response it expects to maintain backwards compatibility.

carneiro-cw: after adding that this and `try_from_alloy_transaction` can be removed.

carneiro-cw (changes requested):

src/eth/types/transaction/transaction_input.rs
```diff
@@ -31,6 +40,28 @@ use crate::eth::types::SignatureComponent;
...
+/// Legacy transaction `v` value constants (EIP-155).
+const LEGACY_V_UNPROTECTED_EVEN: u64 = 27;
+const LEGACY_V_UNPROTECTED_ODD: u64 = 28;
+const EIP155_V_OFFSET: u64 = 35;
+
+impl From<TransactionDecodeError> for alloy_rlp::Error {
+    fn from(value: TransactionDecodeError) -> Self {
+        // Leak the formatted message to satisfy the `&'static str` requirement of `alloy_rlp::Error::Custom`.
+        // This only happens on error paths.
+        let message = Box::leak(value.to_string().into_boxed_str());
```
carneiro-cw: Won't lots of errors just build up this leak indefinetly? Maybe just implement as_str in the error type and return only &'static str there?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -31,6 +40,28 @@ use crate::eth::types::SignatureComponent;
 use crate::eth::types::Wei;
 use crate::ext::RuintExt;
 
+/// Legacy transaction `v` value constants (EIP-155).
+const LEGACY_V_UNPROTECTED_EVEN: u64 = 27;
+const LEGACY_V_UNPROTECTED_ODD: u64 = 28;
+const EIP155_V_OFFSET: u64 = 35;
+
+impl From<TransactionDecodeError> for alloy_rlp::Error {
```
carneiro-cw: Why is this conversion in this file rather than the error definition file?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -31,6 +40,28 @@ use crate::eth::types::SignatureComponent;
...
+    fn from(value: TransactionDecodeError) -> Self {
+        // Leak the formatted message to satisfy the `&'static str` requirement of `alloy_rlp::Error::Custom`.
+        // This only happens on error paths.
+        let message = Box::leak(value.to_string().into_boxed_str());
+        alloy_rlp::Error::Custom(message)
+    }
+}
+
+/// Common fields shared by EIP-2930, EIP-1559, EIP-4844, and EIP-7702 transactions.
+struct AccessListTxFields {
```
carneiro-cw: Why is this called `AccessListTxFields`?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +406,260 @@ impl TransactionInput {
...
+        let input: Bytes;
+        let v: u64;
+        let r: U256;
+        let s: U256;
+
+        match tx_type {
+            TxType::Eip2930 => {
+                gas_price = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("gasPrice"))?;
+                let fields = Self::decode_access_list_fields(&mut rlp, |rlp| {
+                    let to_bytes: AlloyBytes = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("to"))?;
```
carneiro-cw: why are we deconding `to` to alloy types rather than directly to our own?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -143,10 +174,132 @@ impl TransactionInput {
...
+        let mut out = Vec::with_capacity(1 + encoded.len());
+        out.push(tx_type as u8);
+        out.extend_from_slice(&encoded);
+        B256::from(keccak256(out))
+    }
+
+    /// Returns the RLP encoding of the `to` field: empty bytes for contract creation,
+    /// or the 20-byte address for a call.
+    fn encode_to(&self) -> Vec<u8> {
+        self.execution_info.to.map(|addr| addr.0.to_vec()).unwrap_or_default()
```
carneiro-cw, line 215: cant we derive `RlpEncodable` for the primitives as well?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -143,10 +174,132 @@ impl TransactionInput {
...
+        let to = self.encode_to();
+        let input = self.encode_input();
+
+        let tx_type = self
+            .transaction_info
+            .tx_type
+            .and_then(|t| TxType::try_from(t.as_u64()).ok())
+            .unwrap_or(TxType::Legacy);
+
+        match tx_type {
```
carneiro-cw, line 241: I think all of these arms can be collapsed by changing `fn encode_typed_transaction(tx_type: TxType, fields: &[&dyn RlpEncodable])` to `fn encode_transaction(tx_type: Option<TxType>, fields: &[Option<&dyn RlpEncodable>])` and skipping None fields. The field order seems to be the same in each case.

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +406,260 @@ impl TransactionInput {
...
-                transaction_index: None,
-                effective_gas_price: None,
-            })
-            .map_err(|_| rlp::DecoderError::Custom("failed to convert transaction"))?;
-
-            Ok(tx_input)
+impl TransactionInput {
+    /// Decodes the `to` field from RLP bytes: empty means contract creation,
+    /// 20 bytes means a call to that address.
+    fn decode_to(bytes: &[u8]) -> alloy_rlp::Result<Option<Address>> {
```
carneiro-cw: Any reason for the return type not to be `anyhow::Result<Option<Address>, TransactionDecodeError>` ?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +406,260 @@ impl TransactionInput {
...
+            Ok(None)
+        } else {
+            let array = <[u8; 20]>::try_from(bytes).map_err(|_| alloy_rlp::Error::UnexpectedLength)?;
+            Ok(Some(Address::from(array)))
         }
+    }
 
-        let raw_bytes = rlp.as_raw();
+    /// Derives the chain id and signature parity from a legacy `v` value.
+    fn decode_legacy_v(v: u64) -> alloy_rlp::Result<(Option<ChainId>, u64)> {
```
carneiro-cw: Any reason for the return type not to be `anyhow::Result<_, TransactionDecodeError>` ?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +406,260 @@ impl TransactionInput {
...
+        gas_price: u128,
+        gas_limit: Gas,
+        to: Option<Address>,
+        value: Wei,
+        input: Bytes,
+        v: u64,
+        r: U256,
+        s: U256,
+        hash: Hash,
+    ) -> anyhow::Result<Self> {
```
carneiro-cw: Any reason for the return type not to be `anyhow::Result<_, TransactionDecodeError>` ?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +406,260 @@ impl TransactionInput {
...
+        };
+
+        let signer = tx.recover_signer_address()?;
+        tx.execution_info.signer = Signer::Recovered(signer);
+
+        Ok(tx)
+    }
+
+    /// Decodes a legacy transaction from raw RLP bytes.
+    fn decode_legacy(raw_bytes: &[u8]) -> anyhow::Result<Self> {
```
carneiro-cw: Any reason for the return type not to be `anyhow::Result<_, TransactionDecodeError>` ?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +406,260 @@ impl TransactionInput {
...
-                .and_then(convert_tx)
+        let to = Self::decode_to(&to_bytes).map_err(|_| TransactionDecodeError::InvalidTo)?;
+        let hash = Hash::from(keccak256(raw_bytes));
+
+        Ok(Self::build_legacy(nonce, gas_price, gas_limit, to, value, input, v, r, s, hash).map_err(|_| TransactionDecodeError::SignerRecovery)?)
+    }
+
+    /// Decodes the common fields shared by access-list transaction types (EIP-2930, EIP-1559, EIP-4844, EIP-7702).
+    /// The `to` field is decoded via the provided closure because its RLP encoding varies by type.
+    fn decode_access_list_fields<F>(rlp: &mut alloy_rlp::Rlp<'_>, decode_to: F) -> anyhow::Result<AccessListTxFields>
```
carneiro-cw: Any reason for the return type not to be `anyhow::Result<_, TransactionDecodeError>` ?

carneiro-cw: ill stop commenting this on each one, but we don't seem to need to return `anyhow::Error` anywhere

src/eth/types/transaction/transaction_input.rs
carneiro-cw: can't Address implement this directly? Shouldn't the `is_empty` case be dealt with by the caller and this function always return the Address?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +406,260 @@ impl TransactionInput {
...
+    fn decode_legacy(raw_bytes: &[u8]) -> anyhow::Result<Self> {
+        let mut rlp = alloy_rlp::Rlp::new(raw_bytes)?;
+
+        let nonce: Nonce = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("nonce"))?;
+        let gas_price: u128 = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("gasPrice"))?;
+        let gas_limit: Gas = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("gasLimit"))?;
+        let to_bytes: AlloyBytes = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("to"))?;
+        let value: Wei = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("value"))?;
+        let input: Bytes = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("input"))?;
+        let v: u64 = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("v"))?;
```
carneiro-cw: why not decode directly to `U64` which is what we ultimately use in the TransactionInput::SIgnature?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +406,260 @@ impl TransactionInput {
...
+    }
+
+    /// Decodes a legacy transaction from raw RLP bytes.
+    fn decode_legacy(raw_bytes: &[u8]) -> anyhow::Result<Self> {
+        let mut rlp = alloy_rlp::Rlp::new(raw_bytes)?;
+
+        let nonce: Nonce = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("nonce"))?;
+        let gas_price: u128 = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("gasPrice"))?;
+        let gas_limit: Gas = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("gasLimit"))?;
+        let to_bytes: AlloyBytes = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("to"))?;
```
carneiro-cw: why not decode to our primitive `Bytes` type?

src/eth/types/transaction/transaction_input.rs
carneiro-cw: idk if it'd be better though. An alternative would be having our own `TxType` enum and implement an `encode` function that expects the correct params for that tx type.

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +406,260 @@ impl TransactionInput {
...
+                to = fields.to;
+                value = fields.value;
+                input = fields.input;
+                (v, r, s) = Self::decode_signature(&mut rlp)?;
+            }
+
+            TxType::Eip1559 => {
+                gas_price = Self::decode_dynamic_fee_gas_price(&mut rlp)?;
+                let fields = Self::decode_access_list_fields(&mut rlp, |rlp| {
+                    let to_bytes: AlloyBytes = rlp.get_next()?.ok_or(TransactionDecodeError::MissingField("to"))?;
```
carneiro-cw: Genuine question: how does the rlp for bytes differ from a rlp for address? Why can we not decode it directly to address like we do in some other places? Can this field also be something else that is not an address?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +406,260 @@ impl TransactionInput {
...
+        };
+
+        // A raw transaction occupies the entire buffer.
+        *buf = &[];
+        Ok(tx)
+    }
+}
+
+/// Converts a transaction decode error into an `alloy_rlp::Error` for the `RlpDecodable` boundary.
+fn decode_error_to_rlp(error: anyhow::Error) -> alloy_rlp::Error {
```
carneiro-cw: Our decode errors are (or should be) `TransactionDecodeError`, not anyhow::Error.

carneiro-cw (changes requested):

src/eth/types/transaction/transaction_input.rs
f3l1ph3s: this should be the common fields for transactions that have an access list, i will use a more descriptive name

src/eth/types/transaction/transaction_input.rs
f3l1ph3s: nop

src/eth/types/transaction/transaction_input.rs
f3l1ph3s: i was assuming that field could be different for each type of tx but its exactly 0 or 20 bytes for all cases.

src/eth/types/transaction/transaction_input.rs
carneiro-cw: You generalized encode transaction, but did not colapse this match. Any reason why? The match is only required to build the array right?
Something like this should work I think.
```rust

let fields: &[&dyn Trait] = match tx_type {
  TxType::A => &[&a, &b, &c],
  TxType::B => &[&a, &b, &d],
  ...
}
let tx_type = if matches!(tx_type, TxType::Legacy) { None } else {Some(tx_type)}
Self::encode_transaction(tx_type, list)
```

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +422,225 @@ impl TransactionInput {
...
+        let value: Wei;
+        let input: Bytes;
+        let v: U64;
+        let r: U256;
+        let s: U256;
+
+        match tx_type {
+            TxType::Eip2930 => {
+                gas_price = decode_next::<u128>(&mut rlp, "gasPrice")?;
+                let fields = Self::decode_access_list_fields(&mut rlp)?;
```
carneiro-cw: nit: you can do:
```rust
TypedTxCommonFields {gas_limit, to, value, input} = Self::decode_access_list_fields(&mut rlp)?;
```

carneiro-cw (changes requested):

src/eth/types/transaction/transaction_input.rs
carneiro-cw: Doesn't deriving `RlpEncodable` make it so we don't have to call `addr.0.to_vec()`? And instead can just add it to the `&[&RlpEncodable]` array directly?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -143,10 +180,129 @@ impl TransactionInput {
...
+
+    /// Returns the RLP encoding of the `to` field: empty bytes for contract creation,
+    /// or the 20-byte address for a call.
+    fn encode_to(&self) -> Vec<u8> {
+        self.execution_info.to.map(|addr| addr.0.to_vec()).unwrap_or_default()
+    }
+
+    /// Returns the RLP encoding of the transaction `input` data.
+    fn encode_input(&self) -> Vec<u8> {
+        self.execution_info.input.0.to_vec()
```
carneiro-cw: same here

src/eth/types/transaction/transaction_input.rs
```diff
@@ -143,10 +180,129 @@ impl TransactionInput {
...
+    }
+
+    /// Computes the transaction signature hash from the fields stored in this input.
+    ///
+    /// Encodes the unsigned transaction directly via RLP.
+    fn signature_hash(&self) -> B256 {
+        let chain_id = self.execution_info.chain_id.map(|c| c.0.as_u64()).unwrap_or_default();
```
carneiro-cw: same for all of these conversions

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,42 +409,206 @@ impl TransactionInput {
...
+    }
+
+    /// Decodes a legacy transaction from raw RLP bytes.
+    fn decode_legacy(raw_bytes: &[u8]) -> Result<Self, TransactionDecodeError> {
+        let mut rlp = alloy_rlp::Rlp::new(raw_bytes).map_err(|e| TransactionDecodeError::RlpError(e.to_string()))?;
+
+        let nonce = decode_next::<Nonce>(&mut rlp, "nonce")?;
+        let gas_price = decode_next::<u128>(&mut rlp, "gasPrice")?;
+        let gas_limit = decode_next::<Gas>(&mut rlp, "gasLimit")?;
+        let to = decode_to_field(&mut rlp)?;
```
carneiro-cw, line 449: Genuine question, this exists because the field is optional right? Does the crate not implement `RlpDecodable` for `Option<RlpDecodable>`? I think its worth checking if the tx decoding implementation alloy uses treats `to` and other optional fields as special cases

carneiro-cw (changes requested):

src/eth/types/transaction/transaction_input.rs
f3l1ph3s: > Does the crate not implement RlpDecodable for Option<RlpDecodable>?

no afaik

f3l1ph3s: >  I think its worth checking if the tx decoding implementation alloy uses treats to and other optional fields as special cases

right

f3l1ph3s: https://docs.rs/alloy-core/latest/alloy_core/primitives/enum.TxKind.html 🚀

src/eth/types/transaction/transaction_input.rs
```diff
@@ -321,36 +611,40 @@ fn try_from_alloy_transaction(value: alloy_rpc_types_eth::Transaction) -> anyhow
...
 
-    // Recover the signer from the envelope reconstructed using the saved fields.
+    // Recover the signer directly from the saved fields.
     let recovered_signer = tx_input.recover_signer_address()?;
     tx_input.execution_info.signer = Signer::Recovered(recovered_signer);
 
     Ok(tx_input)
 }
 
+fn try_from_alloy_transaction(value: alloy_rpc_types_eth::Transaction) -> anyhow::Result<TransactionInput> {
```
carneiro-cw: Why is this a dedicated function instead of just having `TryFrom<AlloyTransaction> for TransactionInput`?

src/eth/types/transaction/transaction_input.rs
```diff
@@ -302,17 +600,9 @@ impl TryFrom<ExternalTransaction> for TransactionInput {
...
-impl TryFrom<AlloyTransaction> for TransactionInput {
-    type Error = anyhow::Error;
-
-    fn try_from(value: AlloyTransaction) -> anyhow::Result<Self> {
-        try_from_alloy_transaction(value)
-    }
-}
-
-fn try_from_alloy_transaction(value: alloy_rpc_types_eth::Transaction) -> anyhow::Result<TransactionInput> {
+fn build_transaction_input_from_envelope(envelope: &TxEnvelope) -> anyhow::Result<TransactionInput> {
```
carneiro-cw: This too, why not just have the `TryFrom<AlloyTransaction> for TransactionInput`? The only caller is `try_from_alloy_transaction`, and the only caller of `try_from_alloy_transaction` would be `TryFrom::<AlloyTransaction>::try_from`.

src/eth/types/transaction/transaction_input.rs
```diff
@@ -143,10 +177,109 @@ impl TransactionInput {
...
+    ///
+    /// Encodes the unsigned transaction directly via RLP, reusing the alloy RLP
+    /// implementations of our primitive wrapper types.
+    fn signature_hash(&self) -> B256 {
+        let chain_id = self.execution_info.chain_id.unwrap_or_default();
+        let nonce = self.execution_info.nonce;
+        let gas_limit = self.execution_info.gas_limit;
+        let gas_price = self.execution_info.gas_price;
+        let value = self.execution_info.value;
+        let to = TxKind::from(self.execution_info.to.map(AlloyAddress::from));
```
carneiro-cw: I get the idea but we should just implement `RlpDecodable`/`Encodable` for `Option<Address>` IMO. You can copy alloy's implementation for `TxKind`. 

```rust
/// RLP prefix byte for 0-length string.
pub const EMPTY_STRING_CODE: u8 = 0x80;

impl Encodable for Option<Address> {
    fn encode(&self, out: &mut dyn BufMut) {
        match self {
            Some(to) => to.encode(out),
            None => out.put_u8(EMPTY_STRING_CODE),
        }
    }

    fn length(&self) -> usize {
        match self {
            Some(to) => to.length(),
            None => 1, // EMPTY_STRING_CODE is a single byte
        }
    }
}

impl Decodable for Option<Address> {
    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
        if let Some(&first) = buf.first() {
            if first == EMPTY_STRING_CODE {
                buf.advance(1);
                Ok(None)
            } else {
                let addr = <Address as Decodable>::decode(buf)?;
                Ok(Some(addr))
            }
        } else {
            Err(alloy_rlp::Error::InputTooShort)
        }
    }
}
```

src/eth/types/transaction/transaction_input.rs
```diff
@@ -143,10 +177,109 @@ impl TransactionInput {
...
+        };
+        B256::from(keccak256(hash_input))
+    }
+
+    /// Computes the transaction signature hash from the fields stored in this input.
+    ///
+    /// Encodes the unsigned transaction directly via RLP, reusing the alloy RLP
+    /// implementations of our primitive wrapper types.
+    fn signature_hash(&self) -> B256 {
+        let chain_id = self.execution_info.chain_id.unwrap_or_default();
```
carneiro-cw: The bindings here are not needed.

carneiro-cw (changes requested):

src/eth/types/transaction/transaction_input.rs
f3l1ph3s: i think we cant use `impl Encodable for Option<Address>` directly because both are defined outside the crate, `Option<>` is std

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,103 +436,270 @@ impl TransactionInput {
...
-        execution_info: ExecutionInfo {
-            chain_id: value.inner.chain_id().map(Into::into),
-            nonce: Nonce::from(value.inner.nonce()),
-            signer: Signer::Unrecovered,
-            to: match value.inner.kind() {
-                TxKind::Call(addr) => Some(Address::from(addr)),
-                TxKind::Create => None,
+impl TransactionInput {
+    /// Builds a `TransactionInput` from a transaction envelope.
+    fn try_from_envelope(envelope: &TxEnvelope) -> anyhow::Result<Self> {
```
carneiro-cw: nit: move this implementation into `TryFrom<AlloyTransaction>::try_from` and make `TryFrom<ExternalTransaction> for TransactionInput` call `value.0.try_from()`

src/eth/types/transaction/transaction_input.rs
```diff
@@ -31,6 +40,63 @@ use crate::eth::types::SignatureComponent;
...
+
+/// Decodes the next RLP value, mapping a missing value to `MissingField` and an RLP error to `RlpError`.
+fn decode_next<T: RlpDecodable>(rlp: &mut alloy_rlp::Rlp<'_>, field: &'static str) -> Result<T, TransactionDecodeError> {
+    rlp.get_next()
+        .map_err(|e| TransactionDecodeError::RlpError(e.to_string()))?
+        .ok_or(TransactionDecodeError::MissingField(field))
+}
+
+/// Decodes the `to` field: empty bytes mean contract creation, otherwise a 20-byte address.
+fn decode_to_field(rlp: &mut alloy_rlp::Rlp<'_>) -> Result<Option<Address>, TransactionDecodeError> {
```
carneiro-cw: nit: remove this function

src/eth/types/transaction/transaction_input.rs
```diff
@@ -252,103 +436,270 @@ impl TransactionInput {
...
+        *buf = &[];
+        Ok(tx)
     }
 }
 
 // -----------------------------------------------------------------------------
 // Conversion: Other -> Self
 // -----------------------------------------------------------------------------
-impl TryFrom<ExternalTransaction> for TransactionInput {
+impl TryFrom<AlloyTransaction> for TransactionInput {
```
carneiro-cw: This is dead code, can remove

carneiro-cw (changes requested):

carneiro-cw (approved):

PR #2678, opened 2026-09-08
feat: rework configuration from env vars to TOML files

### LINKED ISSUE #2567: Make leader startup config ignore follower config instead of failing
There are lots of configs combinations that could keep the leader from starting, ideally ignore offending configs rather than aborting startup

### LINKED ISSUE #2663: Rework stratus configuration
Currently stratus is configured with env vars. Toml or yaml would be much better, and fit the internal config structure more neatly (lots of configs are nested structs and would map nicely to yaml)

gventino-cw:
Closes #2663

Reworks Stratus configuration from env vars to TOML files. Nested config structs map directly to nested TOML tables, and explicitly provided CLI arguments override file values.

**Precedence:** `defaults < config file < explicit CLI args`

The file is resolved by `--config <path>` or `config/{binary}.{env}.toml` (e.g. `config/stratus.local.toml`). Absent fields fall back to their defaults; unknown fields are ignored with a warning. File values go through the same clap value parsers as CLI values, so invalid values fail at startup with clap's own errors — there is no separate `validate()` pass. See `config/stratus.example.toml` for every option documented.

```toml
leader = true

[executor]
chain_id = 2008

[importer]
external_rpc = "http://localhost:3000/"
sync_interval = "100ms"

[common.tracing]
filter = "info"   # replaces RUST_LOG from env files
```

## Highlights

- Config structs derive clap `Args`; an argument's id is the dotted TOML path of its field, so nested tables map 1:1 to arguments. Durations use humantime strings (`"4m"`).
- File values are converted into `--dotted.id=value` tokens and parsed together with the command line in a **single clap parse** (`args_override_self`), which yields the precedence and the shared validation. A manual scan of `argv` resolves `--config`/`--env` first, since the file must be known before parsing.
- The node mode is explicit: exactly one of `leader`/`follower`/`fake_leader`, from the file or the CLI (required clap `ArgGroup`).
- `executor.chain_id` has no default and rejects zero; follower and fake-leader modes require `importer.external_rpc` (clap `requires` on the mode flags).
- Kafka is all-or-none (`bootstrap_servers`, `topic`, `client_id`) when the connector is created, so a leader tolerates partial sections while a follower fails on them.
- `[common.tracing] filter` replaces `RUST_LOG` from env files (`RUST_LOG` is still honored when the field is absent).
- `config/stratus.local.toml`, `config/stratus-follower.toml` and `config/stratus.example.toml` replace the `.env.local` files; justfile recipes use `--config` and CLI flags.
- Removed: the `CliOverrides` derive from `stratus_macros`, clap `env` attributes, `load_dotenv_file`, `load_env_aliases`, env fallbacks in `stratus_initImporter`, and the `dotenvy` dependency.

## ⚠️ Breaking changes

- All configuration env vars are gone (`CHAIN_ID`, `EXECUTOR_*`, `ROCKS_*`, `KAFKA_*`, `SENTRY_URL`, ...). **Deployments must create the TOML file on the VM before triggering the update on release** (`--config /path/to/stratus.toml`, or the default `config/stratus.{env}.toml` path).
- The node mode and the chain id are now required: a deployment without `leader`/`follower`/`fake_leader` or without a chain id fails to start, with clap naming the missing argument.
- CLI arguments keep working exactly as before and now act as overrides over the file, including partial overrides of `[importer]`/`[kafka]` sections.
- Remaining env vars: `RUST_BACKTRACE`, `RUST_LOG` (fallback), `ADMIN_PASSWORD` (auth secret).

## Test plan

- [x] 30 config unit tests: parsing, precedence, partial overrides, defaults parity, drift coverage (every argument must be coverable by the file)
- [x] 19 `config_cli` integration tests: spawn the real binary and assert the clap-native errors and warnings
- [x] `cargo test --lib` with and without `dev`
- [x] `just lint` clean
- [x] Live smoke tests: leader and follower boot from file with CLI overrides; non-dev builds ignore the dev-only genesis section
- [x] CI
- [x] E2E leader-follower suites

___

### **PR Type**
Enhancement

___

### **Description**
- Migrate configuration to TOML with CLI overrides

- Introduce `CliOverrides` derive for merge logic

- Add config validation for invariants

- Update build scripts and CI for new flags

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["TOML config file"] --> B["Deserialize into structs"]
  B --> C["Apply CLI overrides"]
  C --> D["Validate config invariants"]
  D --> E["Initialize services"]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>16 files</summary><table>
<tr>
  <td><strong>mod.rs</strong><dd><code>Add CommonConfig and StratusConfig definitions</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-b0ba3e54b6f5ef1fa6aed8075e352a68fda66067fbafa8421042a2347801338a">+735/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>cli_overrides.rs</strong><dd><code>Derive macro for CLI override merge logic</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-5b1067a00a0e2c06e4d9667027188bdf8a8ff6c35846afc5ef8dc72d67ccdbac">+311/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>tracing_config.rs</strong><dd><code>Add serde and filter validation to tracing config</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-7925c868efcd8f19bf9e6a02866dd419cd6a35223e1a20991f59b5633b14d2ed">+64/-17</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>mod.rs</strong><dd><code>Enable serde and CLI overrides for permanent storage</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-5c1d69109eb8f0c1ff647aac434f8d48db626a5e159ca8a106ece3e402aef28a">+54/-7</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config.rs</strong><dd><code>Add serde, defaults and CLI merge to RpcServerConfig</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-ebb663b57bd46d8c6a4a6e28823c545b91be35bf5de31e3d593acaf6a58e8b0c">+31/-11</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>config.rs</strong><dd><code>Migrate ImporterConfig to serde with CLI flags</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-0726beefb9e8892b8fd302edfe4f473917dacf46151e3bb93d78d3395e36856e">+35/-10</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>kafka.rs</strong><dd><code>Add CLI overrides and serde to KafkaConfig</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-516e87c7aa8eb8e7908ce3727257a1f4ee49127c5e0e7dcd60fac65c3afcafef">+20/-13</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>config.rs</strong><dd><code>Add serde support and defaults to ExecutorConfig</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-72ec103d3d92cd6b7f97002aecee9877b1d634e1b6878d1768effa91fbf53a64">+45/-8</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>rocks_cf_cache_config.rs</strong><dd><code>Enable serde and CLI overrides for CF cache config</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-bd1af070d1ac4b5f70482f51c3c82bf577c03404e75bdb482887122b215e376b">+3/-10</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>miner_config.rs</strong><dd><code>Add serde, defaults and CLI merge to MinerConfig</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-ea3a8597b909e696993f7e791ab04e5ad784da8077b87088d9c78c804e1ef69d">+30/-6</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>ext.rs</strong><dd><code>Add human-time serde support for durations</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-4ff673578dd06edb71da19750f05cb48bbbd3c78ded3dbeda59409354c35cef6">+42/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>globals.rs</strong><dd><code>Switch to ConfigLoad trait for config parsing</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-48d6fbc3a4ed67100952dcccfada858623c931e73d6de32984d4d1457e68d3a9">+6/-10</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>cache.rs</strong><dd><code>Add serde and CLI overrides to cache config</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-b8200ddbedbd8dab58d6c980baad75e39c70b818c009a4d556ee71b13f046307">+14/-3</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>metrics_config.rs</strong><dd><code>Enable serde and CLI overrides for MetricsConfig</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-82e80998d460f0b1c76d54cdebeeaa6b21f68e1b70d77e4085175a6f3273eae9">+13/-2</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>mod.rs</strong><dd><code>Add serde and CLI merge to StorageConfig</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-5afdbcb51a1480f953cef6a4d26ac8eb7770439927a0ed17ca4316390633a0ac">+5/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>lib.rs</strong><dd><code>Register `CliOverrides` derive in macro lib</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-aa54e2be264adf5d6c46d60aae8c43a6e73f438bbd1425de7653afb7e907ba1c">+6/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Configuration changes</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>loader.rs</strong><dd><code>Implement TOML file loading and CLI merging</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-2f2af8ebb621128659768ef7407f04feab096fa5e9dee1a885dd8265ccd8ee7f">+512/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>justfile</strong><dd><code>Update tasks to use new `--config` and flags</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1">+23/-30</a>&nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>genesis.rs</strong><dd><code>Update tests to use TOML parsing instead of env vars</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-d09e240f407ef1a47fa0bbe3c6f992e78666dd0d47d86dd9f2308d34d877c530">+6/-17</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Bug fix</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>server.rs</strong><dd><code>Remove old env-var importer logic defaults</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-352af3d40d3356fd3836ee6b888f9ca481b0aacc78178fe6334570dabe2a7e9a">+4/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Additional files</strong></td><td><details><summary>12 files</summary><table>
<tr>
  <td><strong>CLAUDE.md</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7">+11/-7</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>Cargo.toml</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542">+2/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>stratus-follower.env.local</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-48ee1351e3ea120c152ce7b7b909a170cd6f0e162aae86477bd71aab2e632ad6">+0/-12</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>stratus-follower.toml</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-7c66df47960d1de28a8098137158695a598959b8285c1ef9dac2123757d15204">+15/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>stratus.env.local</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-c9989f8b14cda489b8b6b3713e7931fdc41cabcabbf3022f26b64a39e9f38ab8">+0/-8</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>stratus.example.toml</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-7d35e1a730ad7f8ea491e7ca04cac064ffddc237b91aefbd479b66f3ed768f19">+147/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>stratus.local.toml</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-0e68aef489caa230659fb2625a0ce10e76dd412ee778c5defa368a2927b2ae77">+11/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>tests.env.local</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-fa4db2065c6452adc59fc5306ab598fac2013aa6bb5432975d9c46fe8d5c792c">+0/-9</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config.rs</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecd">+0/-306</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>mod.rs</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-c652bd8773e9540002a0330f78c15f0474c7b3ef9f85df4e7cda6e09b8951314">+3/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>sentry_config.rs</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-7e0320c0ca7d06025468d37ad64527204635cdb23b573205ac5204ac36d64e5f">+5/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>config.toml</strong></td>
  <td><a href="https://github.com/cloudwalk/stratus/pull/2678/files#diff-331b3f5fea27183313d109f512bc51e8a21ccd8584681b45c448bfa9af8d33ee">+20/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr></tr></tbody></table>

</details>

___


src/config/loader.rs
```diff
@@ -0,0 +1,488 @@
...
+
+/// Configuration that can be loaded from a config file with CLI overrides.
+pub trait ConfigLoad: Sized {
+    /// Loads the configuration, aborting the process with a readable error message on failure.
+    fn load_config() -> Self;
+}
+
+impl ConfigLoad for StratusConfig {
+    #[allow(clippy::expect_used)]
+    fn load_config() -> Self {
```
carneiro-cw, line 49: When I run clippy it doesn't give me a warning here if I remove the allow.

src/config/loader.rs
```diff
@@ -0,0 +1,488 @@
...
+            println!("failed to load configuration | reason={error:?}");
+            std::process::exit(1);
+        })
+    }
+}
+
+impl StratusConfig {
+    /// Loads the configuration: config file as base, explicitly provided CLI arguments as overrides.
+    pub fn load() -> anyhow::Result<Self> {
+        let command = Self::command();
```
carneiro-cw: Just to confirm I got why this is needed:
1. We could get the command using clap's `Args::parse()`
2. After parsing it does not differentiate between defaults and explicitly provided args.
3. Therefore for overrides we need to parse the command ourselves to know only what was provided in the cli.
Is this correct?

crates/stratus_macros/src/cli_overrides.rs
```diff
@@ -0,0 +1,311 @@
...
+        (quote! { #[allow(clippy::unused_self)] }, quote! {})
+    } else {
+        (quote! {}, quote! { #(#statements)* })
+    };
+
+    quote! {
+        #[automatically_derived]
+        impl #struct_name {
+            #lint_allow
+            pub(crate) fn apply_cli_overrides(&mut self, cli: &Self, explicit: &::std::collections::HashSet<::std::string::String>) {
```
carneiro-cw: I think i'd be better if apply_cli_overrides was a function of a Trait rather than a direct implementation in the struct.

src/config/loader.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    /// Loads the configuration from already parsed CLI matches and the given config file content.
+    fn load_from_matches_with_content(command: &Command, matches: &ArgMatches, file_content: &str) -> anyhow::Result<Self> {
+        // parse the config file
+        let mut config: StratusConfig = toml::from_str(file_content).with_context(|| "failed to parse config file".to_string())?;
+
+        // merge explicitly provided CLI arguments over the file values
+        let cli = StratusConfig::from_arg_matches(matches)?;
+        let explicit = explicit_arg_ids(command, matches);
+        config.apply_cli_overrides(&cli, &explicit);
```
carneiro-cw: I don't hate this but I think there might be a simpler path rather than using proc_macros.

At this point we have both the field names and values which we want to override. I feel like we can implement `serde::Deserialize` for `(StratusConfig, overrides)` and apply the overrides during deserialization. 

This might become too verbose though since I think, without implementing a custom `Deserializer`, we'd have to deserialize each field individually.

A more succinct alternative would be

(pseudocode)
```rust
let config: StratusConfig = toml::from_str(file_content).with_context(|| "failed to parse config file".to_string())?;

let overrides = get_overrides();
let config_value: serde_json::Value = config.serialize();

for field, override_value in overrides {
    config_value[field] = override_value
}

let config: StratusConfig = serde_json::from_value();
```

Then we don't need any new macros. I used `serde_json` for serializing here because it returns Value which I know works for this type of thing. The struct is nested though, so it can only really work as shown if we have the full path to the field in the overrides. Still, I think something like this is more straightforward than the macro.

But probably the same could be donse using toml.

src/config/loader.rs
```diff
@@ -0,0 +1,488 @@
...
+        // validate the merged configuration
+        config.validate()?;
+
+        Ok(config)
+    }
+}
+
+/// Resolves the config file path: `--config <path>` when provided, otherwise `config/{binary}.{env}.toml`.
+fn resolve_config_path(matches: &ArgMatches) -> (PathBuf, bool) {
+    if let Some(path) = matches.get_one::<String>("config_path") {
```
carneiro-cw: Maybe config-path should be its own separate clap struct, then we just parse it directly and use serde field markers to define the default path.
This struct could be the actual entrypoint for config parsing in stratus.

src/config/loader.rs
```diff
@@ -0,0 +1,488 @@
...
+        };
+
+        println!("reading config file | path={}", config_path.display());
+        Self::load_from_matches_with_content(command, matches, &file_content)
+    }
+
+    /// Loads the configuration from already parsed CLI matches and the given config file content.
+    fn load_from_matches_with_content(command: &Command, matches: &ArgMatches, file_content: &str) -> anyhow::Result<Self> {
+        // parse the config file
+        let mut config: StratusConfig = toml::from_str(file_content).with_context(|| "failed to parse config file".to_string())?;
```
carneiro-cw: We might need to implement Deserialize for StratusConfig for this, but it would be nice if when parsing a config with non existent fields, we emitted a warning.

src/config/loader.rs
carneiro-cw: Unsure though, the nested nature of the config might make this much more complicated.
The final code generated by the macro is simple enough, reading the actual macro code though is kind of tough.

carneiro-cw: Another alternative to look into would be parsing the toml to Command rather than directly to the config struct, maybe by implementing `Args` or `CommandFactory` (idk which one). We'd use the toml crate to parse it into `toml::Value` (using `.parse::<Table>()`), and use those values to build the Command, using the values that came from the cli to override them instead.
Then clap will deal with validation too (as it does today), instead of us having to implement separate validations.

carneiro-cw: This last one would be by far the best if doable.

carneiro-cw (changes requested):
I think we have to either find a way to parse the TOML to Command and let clap deal with all validations and defaults (best), OR we remove clap completely and rely only on TOML parsing and implement a simple override CLI which integrates more directly with our parsing.

What I'm most worried about is the split of clap defined validations and defaults with TOML parsed values which don't go through clap and thus require separate handling which will most likely lead to unnoticed bugs when updating the config structs.

crates/stratus_macros/src/cli_overrides.rs
gventino-cw: agreed, done

src/config/loader.rs
gventino-cw: it was a leftover from an earlier iteration of this function that still used expect() 

the final version handles the error with `unwrap_or_else + process::exit` so nothing fires, removed

src/config/loader.rs
gventino-cw: yep, that's it

`Args::parse()` returns the struct with clap defaults already materialized, and at that point there's no way to tell a typed value apart from a defaulted one

the distinction only exist in the intermediate `ArgMatches`, where `value_source(id)` reports `ValueSource::CommandLine` vs `ValueSource::DefaultValue`. 

so the loading its split in two steps: 
1. `get_matches()` to collect the explicitly provided ids (`explicit_arg_ids`), and `from_arg_matches()` to materialize the CLI struct

2. the usual "option means not provided" trick doesnt work either, because most args have clap defaults (--address, --executor-chain-id, etc.) and would always look as "provided"

src/config/loader.rs
gventino-cw: no manual `Deserialize` needed, the file now parses through `serde_ignored::deserialize`, which invokes a callback for every field the structs ignore. 

that way `deny_unknown_fields` is gone from all config structs and unknown fields warn instead of failing, smth like this:
 
 `warning: unknown field in config file, ignored | field=executor.chain_idd`

src/config/loader.rs
gventino-cw: done

src/config/loader.rs
gventino-cw: it was doable, via `Command::mut_arg` + `default_value` rather than a `Deserialize` newtype 

`clap 4.5` applies file values as argument defaults at runtime, so no serde `Visitor` work was needed at all

src/config/loader.rs
```diff
@@ -0,0 +1,734 @@
...
+        let command = ConfigCli::command();
+        let matches = command.clone().get_matches();
+        let cli = ConfigCli::from_arg_matches(&matches)?;
+        let (config_path, explicit_path) = cli.resolve_config_path();
+
+        // read the config file, falling back to defaults when the default file does not exist
+        let file_content = match std::fs::read_to_string(&config_path) {
+            Ok(content) => content,
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+                if explicit_path {
```
carneiro-cw: I think resolve_config_path should check for existence and return error, then it doesnt need to return the (string, bool) tuple

src/config/loader.rs
```diff
@@ -0,0 +1,734 @@
...
+
+        // second pass over the same arguments: file values are the defaults now, so clap itself
+        // enforces the precedence and validates file values with the command line value parsers
+        let command = apply_file_defaults(command, &table);
+        let matches = command
+            .try_get_matches_from(std::env::args_os())
+            .context("failed to apply config file values")?;
+
+        config_from_matches(&matches, &table)
+    }
```
carneiro-cw, line 148: This is much nicer than before, but did you try implementing the Deserialization of toml to Command directly?

The idea would be to deserialize toml to Command then pass both the Commands from the cli and the toml to a merge function. The merge function possibly could just be a call to `toml_command.update_from_arg_matches_mut(cli_command)` (cli_command would have to be filtered to only have the explicitly provided args).

This would in theory remove the need for file_only_section.

src/config/loader.rs
```diff
@@ -0,0 +1,734 @@
...
+
+    // flattened `Option<T>` sections only materialize when an argument comes from the command
+    // line: clap marks a group as present just for explicitly provided arguments, so sections
+    // that exist only in the config file are extracted from their applied defaults
+    config.common.sentry = file_only_section(matches, table, "common.sentry", config.common.sentry)?;
+    config.importer = file_only_section(matches, table, "importer", config.importer)?;
+    config.kafka_config = file_only_section(matches, table, "kafka", config.kafka_config)?;
+
+    // validate the merged configuration
+    config.validate()?;
```
carneiro-cw: I think now that we create a command we can leave the validation to clap (i think try_get_matches_from already does this)

carneiro-cw (changes requested):

src/config/loader.rs
gventino-cw: it makes a lot of sense, doing it

gventino-cw: done

src/config/loader.rs
gventino-cw: u're right for value validation: 

* file values are applied as clap defaults, so `try_get_matches_from` runs them through the same value parsers as the command line. The two `validate()` checks that duplicated that are gone (rpc floor, tracing filter, the filter is now rejected by the value parser itself).

what can't move to clap is cross-field validations: 
* its conflict/requirement checks only consider explicitly provided arguments, so file values (which arrive as defaults) are invisible to them. For example:`--leader` + `[importer]` in the file, passes clap and only `validate()` rejects it, `test_cross_field_invariants_between_cli_and_file` covers this cases

src/config/loader.rs
gventino-cw: I did. it was essentially the first iteration of this PR (serde the file into the struct, then overlay the explicit CLI args. the merge API is `Parser::try_update_from` on the derived struct). 

but i've found two blocks while doing that:

- the merge must be filtered to explicitly provided args, otherwise it clobbers file values with clap's built-in defaults: i.e clap materializes defaults into the matches and the derive updater overwrites any field whose id is present. That filter is exactly the `apply_cli_overrides` / `explicit_arg_ids` machinery that was eliminated

- it needs `Deserialize` on every config type, i.e. a second parsing path per value (serde on typed TOML vs clap value parsers on strings like "1s" / "0x2a" / comma-joined lists) with drift risk, today file values go through the very same value parsers as the CLI, just one path to follow it's easier to maintain and keep

all it would save is `file_only_section` (+-10 lines), didn't seem worth it, what u think?

carneiro-cw: > the merge must be filtered to explicitly provided args, otherwise it clobbers file values with clap's built-in defaults

True, but on previous iterations of this PR you successfully filtered out args that were not explicitly set, so really we'd only need to filter the Command that was parsed from the CLI.

> it needs Deserialize on every config type, i.e. a second parsing path per value (serde on typed TOML vs clap value parsers on strings like "1s" / "0x2a" / comma-joined lists) with drift risk, today file values go through the very same value parsers as the CLI, just one path to follow it's easier to maintain and keep

makes sense

> all it would save is file_only_section (+-10 lines), didn't seem worth it, what u think?

would potentially fix having separate validations too.

src/config/loader.rs
carneiro-cw: Maybe in place of `apply_file_defaults` we should convert Table to Command, setting what comes from Table as explicitly set, and then merging the cli command (with only explicitly set flags) on top of it.

Do you think this is feasible (or at least more feasible than implementing Deserialize for Command)?

Something like:

```rust
let file_config = command
       .get_arguments()
       .filter(|arg| !CLI_ONLY_ARGUMENTS.contains(&arg.get_id().as_str()))
       .filter_map(|arg| {
           let value = table_lookup(table, arg.get_id().as_str()).and_then(toml_value_as_string)?;
           let flag = format!("--{}", arg.get_long()?);
           let tokens = match arg.get_action() {
               ArgAction::Set | ArgAction::Append => vec![flag, value],
               _ if value == "true" => vec![flag], // SetTrue flag: valueless token
               _ => vec![],                         // "false": omit
           };
           Some(tokens)
       })
       .flat_map(|tokens| tokens)
       .map(OsString::from);

   // file first, CLI after: args_override_self keeps the LAST occurrence, so the CLI wins
   let merged_config = file_config.chain(std::env::args_os().skip(1)); // skip(1) drops argv[0]

   let mut matches = command
       .clone()
       .args_override_self(true)
       .no_binary_name(true)
       .try_get_matches_from(merged_config)?;

   let config = StratusConfig::from_arg_matches_mut(&mut matches)?;
```

carneiro-cw: Maybe in place of `apply_file_defaults` we should convert Table to Command, setting what comes from Table as explicitly set, and then merging the cli command (with only explicitly set flags) on top of it.

Do you think this is feasible (or at least more feasible than implementing Deserialize for Command)?

Something like:

```rust
let file_config = command
       .get_arguments()
       .filter(|arg| !CLI_ONLY_ARGUMENTS.contains(&arg.get_id().as_str()))
       .filter_map(|arg| {
           let value = table_lookup(table, arg.get_id().as_str()).and_then(toml_value_as_string)?;
           let flag = format!("--{}", arg.get_long()?);
           let tokens = match arg.get_action() {
               ArgAction::Set | ArgAction::Append => vec![flag, value],
               _ if value == "true" => vec![flag], // SetTrue flag: valueless token
               _ => vec![],                         // "false": omit
           };
           Some(tokens)
       })
       .flat_map(|tokens| tokens)
       .map(OsString::from);

   // file first, CLI after: args_override_self keeps the LAST occurrence, so the CLI wins
   let merged_config = file_config.chain(std::env::args_os().skip(1)); // skip(1) drops argv[0]

   let mut matches = command
       .clone()
       .args_override_self(true)
       .no_binary_name(true)
       .try_get_matches_from(merged_config)?;

   let config = StratusConfig::from_arg_matches_mut(&mut matches)?;
```

carneiro-cw: I think this would work on cases such as `--leader + [importer]`

carneiro-cw: because as far as clap is concerned everything is a cli arg

carneiro-cw: Ultimately the decision is up to you, if you think the loader as-is is better we can leave it like it is. Setting the config values as defaults just feels hacky, yk?

gventino-cw: yeah it really feels kinda hacky. I'll try to do it

src/config/loader.rs
gventino-cw: I'll try to do it in a good simple way

crates/stratus_metrics/Cargo.toml
```diff
@@ -11,6 +11,7 @@ clap = { version = "=4.5.60", features = ["derive", "env"] }
 display_json = "=0.2.1"
 serde = { version = "=1.0.228", features = ["derive"] }
 serde_json = "=1.0.149"
+stratus_macros = { path = "../stratus_macros" }
```
carneiro-cw: is this needed still?

src/config/loader.rs
```diff
@@ -0,0 +1,955 @@
...
+            None => Ok(PathBuf::from(format!("config/{}.{}.toml", build_info::binary_name(), self.config.common.env))),
+        }
+    }
+}
+
+impl StratusConfig {
+    /// Loads the configuration: config file as base, explicitly provided CLI arguments as overrides.
+    pub fn load() -> anyhow::Result<Self> {
+        // first pass: resolves the config file path before the file can contribute values
+        let first_pass = command_without_mode_requirement().get_matches();
```
carneiro-cw: If https://github.com/clap-rs/clap/issues/1404 was closed we could just parse and ignore any errors without this function :(.

src/config/loader.rs
```diff
@@ -0,0 +1,955 @@
...
+        let cli = ConfigCli::from_arg_matches(&first_pass)?;
+        let config_path = cli.resolve_config_path()?;
+
+        // read the config file, falling back to defaults when it does not exist
+        let file_content = match std::fs::read_to_string(&config_path) {
```
carneiro-cw, line 115: nit: maybe move this to a func in ConfigCli?

carneiro-cw: everything related to initially parsing the config file really

src/config/loader.rs
```diff
@@ -0,0 +1,955 @@
...
+
+/// Parses the config file content into a TOML table.
+fn parse_config_table(file_content: &str) -> anyhow::Result<Table> {
+    toml::from_str(file_content).context("failed to parse config file")
+}
+
+/// Converts config file values into `--flag=value` command line tokens; arguments the CLI already provides
+/// are skipped, keeping "the CLI overrides the file" uniform.
+fn file_config_tokens(command: &Command, table: &Table, cli_matches: &ArgMatches) -> Vec<OsString> {
+    let cli_provides = |id: &str| cli_matches.value_source(id) == Some(ValueSource::CommandLine);
```
carneiro-cw: doesnt `args_override_self` make this redundant?

src/config/loader.rs
```diff
@@ -0,0 +1,955 @@
...
+    let cli_node_mode = NODE_MODE_ARGUMENTS.iter().any(|mode| cli_provides(mode));
+    command
+        .get_arguments()
+        .filter(|arg| !CLI_ONLY_ARGUMENTS.contains(&arg.get_id().as_str()))
```
carneiro-cw: nit: I think it'd be fine to not filter this. It is a check against a very wierd kind of malformed file that I don't think we need to deal with.

src/config/loader.rs
```diff
@@ -0,0 +1,955 @@
...
+        .filter(|arg| !(cli_node_mode && NODE_MODE_ARGUMENTS.contains(&arg.get_id().as_str())))
```
carneiro-cw: I think `args_override_self` covers this too?

src/config/loader.rs
```diff
@@ -0,0 +1,955 @@
...
+        .filter(|arg| !(matches!(arg.get_action(), ArgAction::Append) && cli_provides(arg.get_id().as_str())))
```
carneiro-cw: why are we filtering args with `ArgAction::Append` ?

src/config/loader.rs
```diff
@@ -0,0 +1,955 @@
...
+        .filter_map(|arg| {
+            let value = table_lookup(table, arg.get_id().as_str()).and_then(toml_value_as_string)?;
+            let flag = format!("--{}", arg.get_long()?);
+            let token = match arg.get_action() {
+                ArgAction::Set | ArgAction::Append => format!("{flag}={value}"),
```
carneiro-cw, line 199: you just filtered out `ArgAction::Append` above?

src/config/loader.rs
```diff
@@ -0,0 +1,955 @@
...
+        .collect();
+    unknown_field_paths(table, "", &known)
+}
+
+/// Recursively collects unknown paths.
+fn unknown_field_paths(table: &Table, prefix: &str, known: &HashSet<String>) -> Vec<String> {
+    let mut unknown = Vec::new();
+    for (key, value) in table {
+        let path = if prefix.is_empty() { key.clone() } else { format!("{prefix}.{key}") };
+        if IGNORED_FILE_SECTIONS.contains(&path.as_str()) {
```
carneiro-cw: In a non-dev build it is actually an unknown field, I think its fine to add it to the vec.

carneiro-cw: And in a dev build it wouldnt have been added to known.

src/config/loader.rs
carneiro-cw: But i still think we need to do something about this. As I understand we had to add defaults to every config value because of this? Honestly manually parsing the args searching for a config path to create a ConfigCli without a StratusConfig seems better to me. 

Would also remove the need for `validate_executor`, `validate_importer` and `validate_kafka` since they wouldn't come with default values and if all fields are not provided clap would error anyway.

carneiro-cw: The bigger issue here seems --help, but then again, you can just implement another struct `ConfigLoader` that only has the path and that loads the toml as args.

carneiro-cw (changes requested):

crates/stratus_metrics/Cargo.toml
gventino-cw: nice catch, I've completly forgot about this

src/config/loader.rs
gventino-cw: did it, the whole first pass now lives in `ConfigCli::parse_first_pass`

src/config/loader.rs
gventino-cw: nope, the mode flags conflict with each other:

* file `--leader` + CLI `--follower` fails with `ArgumentConflict` instead of the CLI winning 

`args_override_self` only lets an argument override itself, skipping the file's mode tokens is what implements the CLI  wins over file behavior from https://github.com/cloudwalk/stratus/issues/2567

src/config/loader.rs
gventino-cw: fair, dropped

src/config/loader.rs
gventino-cw: same as the other convo above, the mode flags conflict pairwise, so both tokens would be an ArgumentConflict, not an override

the skip keeps the CLI mode authoritative over the file's https://github.com/cloudwalk/stratus/issues/2567.

src/config/loader.rs
gventino-cw: because `Append` accumulates occurrences, without the filter the CLI values would extend the file's list instead of replacing it

skipping CLI provided ones keeps "CLI overrides the file" uniform through argument kinds

`test_cli_list_arguments_replace_file_values` pin it

src/config/loader.rs
gventino-cw: only the ones that the CLI provides are filtered out

`Append` arguments coming from the file still reach this match and need the `--flag=value` token form

src/config/loader.rs
gventino-cw: done

src/config/loader.rs
carneiro-cw: Ah true. I don't remember why we have sepparate configs like that. I think it is related with making other configs depend on them. But ok, makes sense

src/config/loader.rs
gventino-cw: did it!

`resolve_config_path` scans argv for `--config`/`--env` without parsing, and the merged parse is now the only clap parse, `--help`/`--version` are handled there directly

`merged_parse_error` and `command_without_mode_requirement` are deleted now, and `executor.chain_id` is a plain required argument

one thing the scan had to absorb it's that token generation still needs to know which arguments the CLI provides (to drop the file's mode tokens and list values), so `cli_provides` answers that from argv by matching each argument's flag. Could not think in a way to this, I'm open to suggestions

I've kept `validate_importer`/`validate_kafka` because these rules are mode-conditional, and clap's `required_if_eq` it doesn't fire on arguments that have defaults. Also dropping the defaults would force `Option<String>` fields through the importer/kafka runtime, and as plain required fields they would break leader mode (I think)

carneiro-cw: > it doesn't fire on arguments that have defaults.

but we can remove the defaults now right? We did not have defaults originally for those, and to me it seems that the only reason to have defaults is to do the first pass. Am i wrong about this?

gventino-cw: removed it and the validate fns too.

src/config/loader.rs
```diff
@@ -0,0 +1,808 @@
...
+        Value::Float(float) => Some(float.to_string()),
+        Value::Boolean(boolean) => Some(boolean.to_string()),
+        Value::Array(items) if items.is_empty() => None,
+        Value::Array(items) => Some(items.iter().map(toml_value_as_string).collect::<Option<Vec<_>>>()?.join(",")),
+        Value::Datetime(_) | Value::Table(_) => None, // not used by any config argument
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
carneiro-cw, line 261: any reason these remaining tests can't go on tests/ ?

carneiro-cw (dismissed):

src/config/loader.rs
gventino-cw: most of them actually can, mb

as u already know tests/ only sees public API, and the merge semantics tests needed an entry point, so I added a small public seam and moved them

carneiro-cw (approved):

